PR #1, opened 2024-07-23
maven build workflow

jodygarnett:
setting up some simple workflows to start out with


jodygarnett:
I am adding this in now just so we have some basic "does it build" protection

PR #2, opened 2024-07-23
Add indexing process

fxprunayre:
Fields removed from the index mapping:
* presentationForm: use cl_…
* hasOverview: can use _exists_:overview query
* hasBoundingPolygon can use _exists_:shape query 
* inspireTheme_syn, inspire_annex, inspireServiceType
* inspireConformResource can be replaced by specificationConformance
* recordOperatedByType can be replaced by linkProtocol
* processRemoteDocs - was never used?


fxprunayre:
Moved branch, I'll reopen.

PR #3, opened 2024-07-24
Adding QA build steps and addressing warnings

jodygarnett:
I am sharing this now as this kind of activity is pretty much ensured to conflict with everyone else. That is why it is best done early in the project.

The interaction between Lombok (which generates code very nicely) and errorprone (which reviews generated byte code happens as plugins to the java compiler. I can get this working in maven fairly well and will need to document the steps for IntelliJ.


src/main/java/org/geonetwork/domain/Group.java
```diff
@@ -68,16 +68,19 @@ public class Group {
...
   @ManyToMany
   @JoinTable(
       name = "group_category",
       joinColumns = @JoinColumn(name = "group_id"),
       inverseJoinColumns = @JoinColumn(name = "category_id"))
   private Set<Category> categories = new LinkedHashSet<>();
 
+  @Builder.Default
   @OneToMany(mappedBy = "iddes")
   private Set<Groupsde> groupsdes = new LinkedHashSet<>();
```
jodygarnett, line 81: @fxprunayre if I understand correctly each time you supply a value like this we need to hint the builder that this should be used as a default value.

src/main/java/org/geonetwork/index/model/record/IndexRecord.java
```diff
@@ -566,7 +569,7 @@ public class IndexRecord {
 
   @JsonIgnore
   @Singular("codelist")
-  private Map<String, ArrayList<Codelist>> codelists = new HashMap<>();
+  final private Map<String, ArrayList<Codelist>> codelists = new HashMap<>();
```
jodygarnett, line 572: However when singular is used the builder is not allowed to do anything, and the field should be marked final.
That is @Builder.Default and @Singular cannot both be used at once...

src/main/java/org/geonetwork/index/model/record/IndexRecord.java
```diff
@@ -647,7 +650,12 @@ public Map<String, Object> getOtherProperties() {
   @JsonAnySetter
   public void handleUnrecognizedField(String name, Object value) {
     try {
-      IndexRecord.class.getDeclaredField(name);
+      Field unexpected = IndexRecord.class.getDeclaredField(name);
+      if (unexpected != null) {
+        System.out.printf(
+          "Handle unrecognized filed %s with value %s for record %s unrecognized conflicts with field %s",
+          name, value, this.getMetadataIdentifier(), unexpected.getName());
+      }
```
jodygarnett, line 658: I understand that this was intended to be executed for a side effect, but am addressing a warning. We may as well cover the case if the field is defined.

src/main/java/org/geonetwork/index/model/record/IndexRecord.java
```diff
@@ -704,7 +714,9 @@ private void handleOperationsProperties(String name, Object value) {
     ArrayList<Integer> operationField =
         operations.computeIfAbsent(name, k -> new ArrayList<Integer>());
     if (value instanceof List) {
-      operationField.addAll((Collection) value);
+      @SuppressWarnings("unchecked")
+      List<Integer> list = (List<Integer>) value;
+      operationField.addAll(list);
```
jodygarnett, line 719: Introduce local variables to address unchecked warnings with relatively minimal impact; and perhaps a slight improvement to readability. This is a real limitation of the Java language...

src/main/java/org/geonetwork/domain/Group.java
fxprunayre: Not really sure, the domain was created using Intellij JPA wizard.

src/main/java/org/geonetwork/index/model/record/IndexRecord.java
fxprunayre: Singular allows to add element one element
Builder provides the default to add the list or map

Screenshot from 2024-07-24 09-09-17

jodygarnett: Yeah I had to revert this back, not sure how to address the warning caused by providing a new map here.

jodygarnett:
@fxprunayre I have removed all changes/fixes to java code so this strictly has the build changes. 
You should be able to rebase cleanly and follow the instructions in docs/README.md

docs/README.md
```diff
@@ -0,0 +1,74 @@
...
+
+```bash
+mvn install
+```
+
+The expectation is that you do a default build before making making a commit or pull request.
+
+The `process-sources` stage is used for code formatting (disable with ``-Dformat.skip=true``):
+
+* [spotless](https://github.com/diffplug/spotless): responsible for ``java`` formatting, using the google google formatting convention
```
fxprunayre, line 32:
```suggestion
* [spotless](https://github.com/diffplug/spotless): responsible for ``java`` formatting, using the google formatting convention
```

jodygarnett:
@fxprunayre I am seeing if I can get the workflows to fail correctly if any formatting was done

PR #16, opened 2024-11-07
better handling of XSLT inside jars

davidblasby:
I noticed that there were issues indexing.

Looking through the problem, the root cause is that the `iso.xsl` and `iso19139.xsl` are package inside JARs.

In some circumstances - especially when running in an IDE - the resources are real filesytem files.  If you are running "normally" (i.e. via maven) these resources will be inside a jar (URLs like `jar:file:path/to/jar.jar!/path/to/resource`).

Since the old code was using `resource.getFile()`, this would work if running exploded (i.e. in the IDE with real filesystem files), but fail if running with the resources inside jars.

This is a very minor fix that, instead of files, uses URLs.  This will work with Filesystem file as well as jar:url.

The test case is a bit complex because of how maven does tests - it will, typically, do them exploded (i.e. resources are real filesystem files).  I tried having tests create a test-jar, however, maven isn't very good at executing them.
So, I put the test cases's .xsl inside the `gn-testing` project.  Then, I use its test-case jar as a dependency for the `gn-utility` module.  

The test case verifies that its a "jar:" url and that `xsl:includes` are working correctly.

### DIFF
--- a/src/shared/testing/pom.xml
+++ b/src/shared/testing/pom.xml
@@ -48,7 +48,6 @@
         <artifactId>maven-jar-plugin</artifactId>
         <executions>
           <execution>
-            <phase>validate</phase>
             <goals>
               <goal>test-jar</goal>
             </goals>
--- a/src/shared/testing/pom.xml
+++ b/src/shared/testing/pom.xml
@@ -48,6 +48,7 @@
         <artifactId>maven-jar-plugin</artifactId>
         <executions>
           <execution>
+            <phase>compile</phase>
             <goals>
               <goal>test-jar</goal>
             </goals>
--- a/src/shared/testing/pom.xml
+++ b/src/shared/testing/pom.xml
@@ -54,6 +54,10 @@
             </goals>
           </execution>
         </executions>
+        <configuration>
+          <forceCreation>true</forceCreation>
+          <skip>false</skip>
+        </configuration>
       </plugin>
     </plugins>
   </build>
--- a/src/shared/testing/pom.xml
+++ b/src/shared/testing/pom.xml
@@ -48,7 +48,7 @@
         <artifactId>maven-jar-plugin</artifactId>
         <executions>
           <execution>
-            <phase>compile</phase>
+            <phase>validate</phase>
             <goals>
               <goal>test-jar</goal>
             </goals>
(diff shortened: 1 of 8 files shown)


davidblasby:
This does use a StreamSource (via url).  The issue was it was, previously, calling resource#getFile() - if it was a `jar:` resource, this would return null (causing an issue).  Just removing the #getFile() call fixed it - easy fix.

The hard part, for this, was the test case...

fxprunayre:
Indeed, in GN4, all XSLTs are in filesystem. So we have to decide if better to pack them in JAR or keep the current way.

Also GN use a resolver with https://github.com/geonetwork/geonetwork/blob/main/src/shared/schemas/src/main/resources/xml/resolver/oasis-catalog.xml which requires to unpack resources in JAR in a folder - for now done in https://github.com/geonetwork/geonetwork/blob/main/src/shared/schemas/src/main/java/org/geonetwork/schemas/SchemaManager.java#L191-L198. I'm not sure we can avoid that?

davidblasby:
It appears that saxon is fine with `jar:...` URLs and following them relatively for processing `xsl:include`s.  The test case verifies this.

The test case puts two `xsl` documents into a jar, and then executes.  The main document uses `xsl:include` to load in the 2nd document.  The test case verifies that the execution was correct.

I don't think there will be any issues with XSL documents in jars - but definitely something to keep an eye on.  I think the resolver will be good for security - to make sure that a malicious XSL isn't resolving local `http:` resources.    But, I think security is a bigger concern.

However, I think there were some GN4 assumptions about a lot of the "files" being actual filesystem resources (instead of URLs).  Not sure if that is necessary for GN5 (unlikely), or just something to make life easier developing GN4 (much easier to debug a filesystem file instead of a file inside a jar...).

davidblasby:
I will merge this at the end of the week @fxprunayre @josegar74

src/shared/testing/pom.xml
```diff
@@ -53,6 +53,10 @@
             </goals>
           </execution>
         </executions>
+        <configuration>
+          <forceCreation>true</forceCreation>
+          <skip>false</skip>
+        </configuration>
```
jodygarnett, line 59: That is interesting so you have it always build the test jar....

jodygarnett (approved):

fxprunayre:
Similar issue when using a StreamSource, setting setSystemId with the URL fix the include resolution (https://github.com/geonetwork/geonetwork/pull/78/commits/18b085553e8891640812317d8c84daf14c4decd2)

Issue #50, opened 2024-11-14
Records


fxprunayre:
* #66 
* Editing (#73)
* Sharing
* Validation (XSD, Schematron, Shacl, ...)
* #67 
* #68 
* Restricted section with with held attribute support
* Directory (for contact, ...)
* ...

Current API: https://apps.titellus.net/geonetwork/doc/api/index.html#/records

## Questions

* Sharing / A more simple publish/unpublish only mode?
  * Some user use only the workflow to manage publication state
  * https://github.com/geonetwork/core-geonetwork/issues/8637
* Model / Existing schema plugin usage?
* Model / Keep schema plugin capability?
* Model / Keep XML encoding?
* Editing / Keep capability to parse XSD and create meta document? 
* Editing / How to edit? (currently form generated using XSLT, or batch edit operations or XML import)
* Editing / Batch editing 
  * how? GeoNetwork users have an history or records - How to facilitate batch work?
  * Deprecate batch editing config https://github.com/geonetwork/core-geonetwork/blob/main/schemas/iso19115-3.2018/src/main/plugin/iso19115-3.2018/layout/config-editor.xml#L951 - not really usable ?
  * Add a new mode based on the Index model property to be able edit with well known property (replacing XPath and XML snippet) eg. https://github.com/geonetwork/geonetwork/blob/features/frontend-exp-indexrecord/src/modules/editing/src/test/java/org/geonetwork/editing/BatchEditsControllerIntegrationTest.java#L129
* Attachements
  * Add options to not embed the content of the datastore in MEF (we sometimes have Go of data) 

## Actions

- [ ] Migrate API

jahow:
Here is what I think could work in GeoNetwork 5 on this topic:

# Editing

## No XSLT for manipulating records

As demonstrated in GeoNetwork-UI creating and modifying an XML document in the browser is entirely feasible, faster and more reliable (tested).

Currently in GeoNetwork 4 each schema plugin needs to define a lot of operations in XSLT:
* modifying online resources
* modifying thumbnails
* modifying links
* cleaning up
* update fixed info

Leaving the responsibility of manipulating the XML to the client means simplifying the schema plugins.

## API for editing

GeoNetwork 5 would only provide a simple CRUD API for records. Note that validation should still be possible but that is another topic (see below).

## Batch editing

The GeoNetwork backend should still be able to operate on documents in batches. I don't have a technical solution in mind right now but I'm confident that we can find one without delegating this responsibility to schemas.

[...]

PR #73, opened 2024-11-22
API / Editing

fxprunayre:
Draft migration of editing API

image

- [x] Build metadocument
- [x] Add element
- [x] Update element
- [x] Delete element
- [x] Save
- [ ] Cancel
- [ ] Add attribute
- [ ] Remove attribute
- [ ] Reorder element
- [ ] Form builder
- [ ] ...

Number of side tasks to discuss:
* Validation
* Record history
* Workflow
* With held elements
* XLinks
* AccessManager
* ...

Q:
* To discuss if we simplify MetadataUtils, XmlSerializer and what would be good for "invasive" features like workflow
* API doc / Any ideas/guideline to improve API doc? More details, more examples?
* API / Expose meta document for advanced client app willing to use it? (in GN4, only the HTML form generated from it is exposed. It is also a good document to look at to understand how editing works.)
* Form / Approach to be define? We can probably reduce XSLT use.
* Would be good to define the Exception strategy

### DIFF
--- a/src/modules/editing/pom.xml
+++ b/src/modules/editing/pom.xml
@@ -61,7 +61,17 @@
     </dependency>
     <dependency>
       <groupId>org.geonetwork</groupId>
+      <artifactId>gn-indexing</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.geonetwork</groupId>
       <artifactId>gn-constants</artifactId>
+      <version>${project.version}</version>
+    </dependency>
+    <dependency>
+      <groupId>org.geonetwork</groupId>
+      <artifactId>gn-legacy-utility</artifactId>
       <version>${project.version}</version>
     </dependency>
     <dependency>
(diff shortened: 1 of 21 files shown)


jahow:
Hi, what is the status on this PR? I don't think porting back the whole editing system from GN4 was approved by a proposal. I wouldn't be in favor of it, as I think editing metadata can be delegated to the client app (which is done and works reliably in geonetwork-ui).

jahow:
To be a bit more exhaustive: I have the feeling that doing editing in XML/XSLT induces a lot of complexity downstream, especially for schema plugins. This should be discussed IMO

fxprunayre:
> I don't think porting back the whole editing system
> I have the feeling that doing editing in XML/XSLT induces a lot of complexity downstream, especially for schema plugins. 

Note that the editing API is mainly the capability of GeoNetwork to parse XSD and create meta documents with domain knowledge. Then the editing form can be using this API and be generated using XSLT or anything else.

> This should be discussed IMO

Added some questions here with also consideration about batch editing:
https://github.com/geonetwork/geonetwork/issues/50
Don't hesitate to add more.

PR #74, opened 2024-11-26
OGCAPI-Records for GN5

davidblasby:
I'm adding this early so people can see what I am working on.

- [x] Initial work getting the OGC's OpenAPI (.yaml) working
- [x] Initial test cases for the model
- [x] Implementation of all the services
- [x] Testing 

[Summary PPT](https://github.com/user-attachments/files/18443641/gn5-ogc-records.pdf)

Please see the included .YAMLs.  These are originally from the [ogc's ogcapi-records github](https://github.com/opengeospatial/ogcapi-records/tree/master/core/openapi) repo.

I've modified them so they will produce "good" java code. There are quite a few minor changes - see the `README.md` in some of the directories.  Most of the changes I made don't affect the API.  

However, a few do make "real" API changes - mostly on the output side.  For example;

* IDs are always Strings (in the spec they can be a number or string).  For parsing the .json, numbers will automatically be converted to strings
* `catalog#recordType` is always a `List<String>` (in the spec they can be a single string or string array)
* I've modified how the GeoJSON parsing happens.  This doesn't affect the model, however, you must ensure that you set the sub-class (i.e. `OgcApiRecordsPolygonDto.java`) `type` (it not automatically done).

More details are in the `README.md` and comments at the top of the `.yaml` files.

I've added test cases for the seralizing/deserializing:
* examples from the OGCAPI spec
* examples that test some of the changes I made to the model


davidblasby:
@josegar74 @fxprunayre - please take a look. 

I will be looking at doing some actual implementations next.

davidblasby:
@fxprunayre @josegar74 @juanluisrp - This is getting to the point of being usable now.
I just added documentation on how to setup a multi-portal (multi-collection) with "extra" collection metadata.

I havent done much testing, but it appears to be working.  

Please review.  I've asked @juanluisrp to make a deployment.

davidblasby:
Added PDF/PPT summary in the description that goes through a lot of technical details - also, good info about GN5.

davidblasby:
Ok, I've tested this against a few clients.  Each of the clients has a few "quirks" - I updated GN5 OGCAPI-Records to account for this.

1. GeoCat's OGCAPI-Records (originally for geonetwork-microservices)
2. QGIS - MetaSearch plugin.  Use a specific collection as the url (i.e. http://localhost:7979/ogcapi-records/collections/subportal)
3. STAC Browser -  https://radiantearth.github.io/stac-browser 
4. Camp2Camp Info - https://camptocamp.github.io/ogc-client/#/demo (you need to host GN5 with HTTPS or use something like http://ngrok.com)
5. Open Geo Groep -  https://opengeogroep.github.io/ogc-api-features-testclient/src/index.html

davidblasby:
I noticed that all the clients (except my Angular) do not handle queryables.

I've added support for queryables in the `q` (query string).

http://localhost:7979/ogcapi-records/collections/subportal/items?q=contacts:geocat%20portal

This will search for "portal" with the `contacts=geocat` queryable.

This way you can use the queryables anywhere.

davidblasby:
added the elastic IndexRecord to catalog (i.e. `landing page` and `/collections` and `/collections/<collectionid>` and the individual records). 

I've also added this into the OpenAPI Swagger document, and it should keep itself updated when `IndexRecord.java` is modified (will need to build `src/module/ogcapi-records`).

davidblasby:
updated documentation

davidblasby:
@josegar74 @fxprunayre - will merge this monday.

docs/manual/docs/OGCAPI-Records/develop.md
```diff
@@ -0,0 +1,110 @@
+# Technical Overview
+
+Here is a technical overview so developers can easily contribute!
+
+## History
+
+The [GeoServer Microservices Project](https://github.com/geonetwork/geonetwork-microservices) had an OGCAPI-Records implementation.  This was used as a basis for the GN5 implementation.
```
fxprunayre, line 7:
```suggestion
The [GeoNetwork Microservices Project](https://github.com/geonetwork/geonetwork-microservices) had an OGCAPI-Records implementation.  This was used as a basis for the GN5 implementation.
```

davidblasby:
@fxprunayre - thanks for the geoserver catch!  Oops!

davidblasby:
@fxprunayre @josegar74  - I'd like to merge this ASAP.  Its bit rotting and making other work more difficult.

fxprunayre:
Tested on a quite large catalog and works @davidblasby but I noticed some issues:
* on items
```
{"code":"500","description":"Cannot invoke \"java.time.OffsetDateTime.toString()\" because \"x\" is null"}
```
but with no details in the log so hard to track where this comes from. You can double check https://github.com/davidblasby/geonetwork/commit/6ce65d712fba180767c6e8b4db452bfc91a9d3ff


* We may have to support UUID with special characters

```
{"code":"500","description":"java.net.URISyntaxException: Illegal character in path at index 113: http:/localhost:7979/ogcapi-records/collections/b08fe709-1ced-4a07-8edf-06aa6ccdf2e3/items/SDN_CPRD_1568_KN172L06 SeaBeam 3012-P1 Bathymetry Grid?f=geojson"}
```

* Requesting large response, I'm getting

```
{"code":"500","description":"node: http://localhost:9200/, status: 200, [es/search] Failed to decode response"}
```

* in http://localhost:7979/doc/api/swagger-ui/index.html the server URL is wrong

image



Is there any way to get exception details?

david-blasby:
Hi, Francois,

There is code to handle the exception in `src/modules/ogcapi-records/ogcapi-records-implementation/src/main/java/org/geonetwork/ogcapi/controllerexceptions/GlobalExceptionHandler.java`.  We should probably have it log the error (and stacktrace).

The special chars (like ' ') in the UUID probably just needs to make sure that we URL-encode the links.  How are you getting to that page?  If you are copy-and-pasting it to the browser location bar, then we probably need to make sure that spring boot will automatically fix that url before internal dispatch. 
However, if you are getting that link in the JSON, then we should just make sure that we are URL-encoding the links.

I will look at this tomorrow:

1. log exceptions
2. URL-Encode links
3. test and see what happens when you have `space` chars in the browser URL-bar.
4. look at the swagger UI output.  I expect this has something to do with how swagger is building the UI based on the various endpoints.  When I looked at the json output, i didn't see these problems...

Thanks for your feedback,
Dave

david-blasby:
pushed commit for;

a) turning on controller error logging
b) better url encoding of uuid/collectionid
c) problem was in the link generation - not in the controller endpoints

 @fxprunayre  - can you give me some context for the "Requesting large response, I'm getting".  What are you doing?  Can you give me the exception stack trace (and caused by).  You should be able to get it easily with the latest commit.

I will look at the swagger stuff tomorrow.

davidblasby:
The swagger stuff was a quick fix - not sure why it was a problem...

@fxprunayre - can you give me some context for the "Requesting large response, I'm getting". What are you doing to cause it? Can you give me the exception stack trace (and caused by)? You should be able to get it fairly easily with yesterday's erorr-logging commit.

davidblasby:
@fxprunayre  - I did the merge commit inside github because it "looked easy."  Turns out it was more complex. 

FIXED.

davidblasby:
@fxprunayre @josegar74 - I would like to merge this.

src/modules/ogcapi-records/ogcapi-records-implementation/src/main/java/org/geonetwork/ogcapi/configuration/WebConfig.java
```diff
@@ -0,0 +1,74 @@
...
+import org.springframework.web.servlet.config.annotation.ContentNegotiationConfigurer;
+import org.springframework.web.servlet.config.annotation.WebMvcConfigurer;
+
+/**
+ * configuration for the Spring Boot app.
+ *
+ * <p>This contains some beans and some overrides for setting up spring.
+ */
+@Configuration
+public class WebConfig implements WebMvcConfigurer {
```
PascalLike, line 32: Did you consider the option of using the annotation @EnableWebMvc?
 https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/servlet/config/annotation/WebMvcConfigurer.html

davidblasby: yes - this was the recommended way to setup the ContentNegotiationConfigurer and provide an objectmapper.

Is there a better way?

davidblasby: I am going to merge this - but its actively being worked on.  If you have a better way, I can change it.

src/shared/ogcapi-records/src/main/java/org/geonetwork/ogcapi/service/validation/SimpleValidators.java
```diff
@@ -0,0 +1,29 @@
...
+     * validates a catalogId. This must only have letters, numbers, or "-" or "_".
+     *
+     * @param catalogId from user
+     * @throws Exception if the catalogId is bad
+     */
+    public static void validateCatalogId(String catalogId) throws Exception {
+        if (isBlank(catalogId)) {
+            throw new Exception("no catalogId!");
+        }
+        if (!catalogId.matches("^[a-zA-Z0-9_-]+$")) {
```
fxprunayre, line 25: This is a bit more strict than what GN4 does. We may end-up with message like

```
{"code":"500","description":"java.lang.Exception: catalogId 'metawal.wallonie.be' is invalid!"}
```
when the main catalogue id is set in the settings:

Image



For portals, we have a client side check on portal name

Image

but none on API or DB side.

Should we be more strict in all places?

PascalLike: These restrictions are not from the OGC API specification directly, but from the fact that collectionId (as well as itemId, which are metadata identifiers) are used in URLs. Generally, it would be sufficient to use characters that can be used in URLs... Then I think the period `.` could be added to the regex. What do you think @davidblasby ?

fxprunayre: It makes me think of this old discussion in GN4 also https://github.com/geonetwork/core-geonetwork/pull/5736

In OGC API spec `collectionId` is `string` or there are some constraints?

davidblasby: I found that accepting anything as an ID would be problematic from a security perspective.

Feel free to open this up to a few more characters (like ".").  

Records can have some craziness for their IDs.

Here is what google-ai recommends:

```
Best Practices for "Safe" IDs

To ensure your API works across all clients and servers, follow these "unwritten" rules:

- Stick to URL-Safe Characters: Use a-z, 0-9, -, and _. Avoid spaces and special symbols.

- [Kebab-case or Snake_case: These are standard for {collectionId} (e.g., natural-parks or water_wells).

- UUIDs for Features: If your data doesn't have natural IDs, UUIDs (e.g., 550e8400-e2...) are a safe, standard way to ensure uniqueness across systems.

- Avoid Lengthy IDs: Extremely long IDs can sometimes hit URL character limits in older browsers or proxy servers. 
```

Issue #66, opened 2024-11-14
Formatters


fxprunayre:
Current API: https://apps.titellus.net/geonetwork/doc/api/index.html#/records/getRecordFormattedBy

## Questions

- Need cache ? Using Spring ?
- Some project added possibilty to save formatter results in file system (eg. in HTML) for landing page (for better SEO indexing)

## Actions

- [ ] Migrate API
- [ ] Reuse formatters in OGC API Records

## Related work

davidblasby:
With emphasis on the new GN4 DCAT formatters (XSLT).

PR #78, opened 2024-12-12
API / Formatter migration check

fxprunayre:
DO NOT MERGE

This is superseded by - https://github.com/geonetwork/geonetwork/pull/96

Draft evaluation on how to migrate the formatter API.

One goal is also to be able to use formatters in OGC API Records.

In GN4, we can distinguish 3 kind of XSLT formatter:
1. XSLT having Java function call (eg. DCAT)
1. XSLT having `xsl:include` to resources outside the formatter folder (eg. ISO19139 DCAT, xsl-view)
    * XSLT in convert folder
    * XSLT in layout folder
    * XSLT in formatter from another plugin
1. XSLT without above dependencies (eg. datacite)

Migration steps:
* For 1), migrate Java functions to package `org.geonetwork.utility.xml.xslt`
* For 3), copy/paste
* For 2), ...

Currently tested:

* http://localhost:7979/api/records/d82a7ade-a1f0-4862-9a99-78006751e631/formatters/xml = XML document from the database
* http://localhost:7979/api/records/d82a7ade-a1f0-4862-9a99-78006751e631/formatters/json = JSON document from the index
* http://localhost:7979/api/records/d82a7ade-a1f0-4862-9a99-78006751e631/formatters/datacite = Standalone XSLT formatter
* http://localhost:7979/api/records/d82a7ade-a1f0-4862-9a99-78006751e631/formatters/dcat = (ISO19115-3) XSLT formatter with Java function call 

Additional items supported in GN4, to be discussed if need to be ported or not, as-is or not, ...

- [ ] Parameters to review eg. `mdPath`, `width`
- [ ] `withHeld` support
- [ ] Workflow
- [ ] PDF support
- [ ] Caching

Related to #66


jahow:
Hi,

I think this needs to be discussed further before any real implementation is approved. The concept of formatters in GN4 was limiting because of the absence of a pivot format, which meant that the schemas that could be used for storing metadata was de facto very limited. The geonetwork-microservices OGC API Records service introduced outputting a list of records by using the index as a source of truth, and I think this is a very promising approach going forward. Using this same principle to output DCAT, CSW etc. should be seriously considered in my opinion, in order to avoid having many redundant code paths.

davidblasby:
@fxprunayre - I have been trying this out.

1. Do you have some sample records?  The only `iso 19115-3.2018` record (required for dcat) I could easily find was the single one in the GN schema sample (add via the gn4 admin screen).

2. I was able to get it to output this record (`0dbac056-cd5a-4c9c-bbec-564e47c4507c`) to dcat!  🥇 

4. I had quite a few problems with security (`MetadataAccessManager`).  I've added more support to this and made some changes to the security system.  I will make a PR for you, tomorrow (starting your thursday evening), and we can discuss.

Thanks for putting this together.  

What are the next steps?  Should we start moving all the formatters from GN4 to GN5?  Is this something difficult to do (i.e. no java calls).

jahow:
Hey @davidblasby! Please keep in mind that this topic needs to be discussed by the community and approved by a PSC before anything gets merged. Formatting is an integral part of GeoNetwork and the architecture choices made here will have very important consequences.

I wouldn't recommend investing too much time on implementation before such discussions have happened.

fxprunayre:
> * Do you have some sample records?  The only `iso 19115-3.2018` record (required for dcat) I could easily find was the single one in the GN schema sample (add via the gn4 admin screen).

https://metawal.wallonie.be/geonetwork is in ISO19115-3

ticheler:
> Hi,
> 
> I think this needs to be discussed further before any real implementation is approved. The concept of formatters in GN4 was limiting because of the absence of a pivot format, which meant that the schemas that could be used for storing metadata was de facto very limited. The geonetwork-microservices OGC API Records service introduced outputting a list of records by using the index as a source of truth, and I think this is a very promising approach going forward. Using this same principle to output DCAT, CSW etc. should be seriously considered in my opinion, in order to avoid having many redundant code paths.

Hi @jahow,
At this point in time, having a pivot format as single point of truth is not something we can implement. There are to many uncertainties around that, since it will not allow us to use the official validation rules (schema, schematron for many different country or topic specific profiles) that will ensure the source metadata record remains valid and identically structured. 
Also, there is not yet a JSON encoded version of ISO19115, although that is in the works. It will take considerable time before that is an accepted standard. There is also work on official mapping between ISO19115 and (Geo)DCAT that is not formalised yet. All of these things and other considerations make that we can not move the point of truth from the original XML encoded and validated records to the JSON encoded index. I'm not saying it can't happen in the future, but it is not an option at this point in time.
At the same time, I'm open to your contributions of a native DCAT support in GN5 that takes into account our current requirement to maintain XML documents as point of truth for many of the supported metadata records.

jahow:
Thanks for the feedback @ticheler. Could you please let us know if there is indeed a plan to write a proposal as a basis of discussion for this topic?

> At this point in time, having a pivot format as single point of truth is not something we can implement. There are to many uncertainties around that, since it will not allow us to use the official validation rules (schema, schematron for many different country or topic specific profiles) that will ensure the source metadata record remains valid and identically structured.

I'm not sure I follow you. In GN4 the index _is_ a pivot format, and it was used to produce the OGC API Records output. I'm simply suggesting that we go with a similar direction for things like DCAT and others. Having such a format does not mean that we discard the original XML document, on which we can still run validation checks.

davidblasby:
Hi, Olivia, Thanks for your comments and concerns!

The GN5 version of OGCAPI-Records (PR 74) does, mostly, use the Elastic JSON Index record (in the same manner as `GN-microservices`).  However, I think `GN-microservices` used XSLT directly on the XML record (not 100% sure) for HTML.  Currently, in GN5, i've added the Elastic JSON record and the underlying record (i.e. XML) so a client can choose how they want to render output.

The goal of moving the `FormatterApi` over (from GN4) is to get access, in particular, for the new DCAT-* that @fxprunayre and @josegar74 have been working on (see several recent GN4 PRs).  The conversions are defined (by the standard's body) from the underlying ISO XML to the DCAT-* format and not from a pivot model...  I expect someone in the standards body is  doing some work on a pivot model - but I'm not sure what its state is.  I talked to a few people doing this and they sounded hopeful, but also extremely cautious about its feasibility - but I don't want to speak on their behalf! 

I think there will always be interest in conversion from the underlying ISO XML to new standards.  I'd, also, like to leverage the DCAT-* work that Francois and Jose have been doing in GN4 inside GN5 without having a lot of maintenance overhead.  I am looking forward to adding this to the GN5 OGCAPI-Records and have it available to other modules. There is a lot of excitement (especially in the EU) about DCAT-*.

I expect that GN4 and GN5 will be working together for quite a while - and I think that's a good thing.  However, it does mean that radical departures really need to be thought out and planned accordantly.  

The actual development community is quite small so it important to not take on too much, have achievable goals, evolve the technical infrastructure, and evolve existing deployments:

https://github.com/geonetwork/core-geonetwork/graphs/contributors?from=2%2F25%2F2023
https://github.com/geonetwork/geonetwork/graphs/contributors?from=2%2F24%2F2024

I don't have any problem with having something like the Elastic JSON index having a more prominent role in GN5 and have less reliance on schema-specific XSLT.  That's been successful for the OGCAPI-Records (both in GN5 and in geonetwork-microservices).  Both of these projects have also greatly improved the Elastic Indexing model (including a few changes inside GN4 as well).  However, as mentioned above, there's always going to be demand for XSLT - including from standards bodies.  I am, also, hoping, that GN5 has significantly less reliance on XSLT than GN4.

There's also room for conversions from the pivot model (i.e. Elastic JSON) to other formats and no one is closing the door on this.  In fact, this is the core of the OGCAPI-Records and it would be great to make this easier and have more technical support!

I believe that @ticheler will be meeting with some of the C2C folks March 10th to discuss.

ticheler:
> Hey @davidblasby! Please keep in mind that this topic needs to be discussed by the community and approved by a PSC before anything gets merged. Formatting is an integral part of GeoNetwork and the architecture choices made here will have very important consequences.
> 
> I wouldn't recommend investing too much time on implementation before such discussions have happened.

Not sure how you envision this to proceed. At this point in time, 100% of the work on GN5 is done by a group of developers that have a very good insight in the technical requirements and challenges that they face. The current PR will add a formatter that works to the project, allowing to use the point of truth metadata to be transformed into required and properly mapped output. It does not prevent to also use the json indexed version of those records.
Again, feel free to come up with suggestions and code that ensures a 100% support for validated ISO19x compliant metadata records and we can consider to work with you on that.

ticheler:
> I'm not sure I follow you. In GN4 the index _is_ a pivot format, and it was used to produce the OGC API Records output. I'm simply suggesting that we go with a similar direction for things like DCAT and others. Having such a format does not mean that we discard the original XML document, on which we can still run validation checks.

Indeed that will be a path, it is a path that can be used, and it should not stop this formatter PR and further progress.

jahow:
Oh want, thanks for all the responses. It doesn't sound like you folks are planning a proposal stage, am I right in saying that? The original agreement on GN5 was that there would be additional proposals for each of the modules that would be migrated to GN5.

I think I understand all the arguments presented here, but I still have the feeling that this approach of formatting 1/ relies extensively on XSLT (which we agreed on reducing instead) 2/ brings back a lot of the technical quirks of GN4 that were there for historical reasons (formatting both from the index and from the source XML).

Also, what's the point of doing the same as in GN4 if we can still delegate to GN4 the stuff we need in the meantime? I don't see the rush there.

I'll gladly work on implementing DCAT formatters from a pivot format as soon as C2C gives me some time to work on this.

davidblasby:
Hi, Oliva,

The XSLT for DCAT-*  conversion is based on the standards bodies.  In fact, I believe some of these conversions are defined/specified in XSLT and there are official XSLTs published (but not for all).  It is already in GN4 (see francois and jose's recent PRs), including some country specific outputs.  I think there are plans to improve and expand upon these.

> Also, what's the point of doing the same as in GN4 if we can still delegate to GN4 the stuff we need in the meantime? I don't see the rush there.

As I mentioned, I'd like to use the DCAT (and hopefully other) format conversion that are already in GN4 in GN5 for the OGCAPI-Records.  I'd like to share maintainance and keep the same XSLT in GN4 and GN5 - I certainly don't want to be responsible for a totally different way of doing that in GN5!  I like how GN5 is currently setup and I can easily bring in francois' work, and I appreciate him doing this in public and being available to help with it!  Conversions (esp DCAT) are very important, and I am very glad he's put in the work now to get them working and easily re-usable!  I'd like to get this into OGCAPI-Records in March/April - not 2026/2027.  The OGC and Metadata groups are really excited by this development and watching it.

> I'll gladly work on implementing DCAT formatters from a pivot format as soon as C2C gives me some time to work on this.

As I mentioned, there are some official DCAT conversions in the specification space and there is some work on metadata "pivot formats".  The country-specific DCAT-* tend to follow along with these.  I havent had the time to go into them in the extreme depth required - I would love to hear what you learn from them!

I'd, also, love to see an easier-to-use plugin system (and formatter metadata) for the `FormatterApi` - I can see XSLT as well as "from pivot format" support.  I think that's a longer-term goal.

jahow:
Hey all,

I appreciate the profuse feedback :) shouldn't we organize a meeting to discuss this in a easier way? Because I feel like there's a lot of misunderstanding here.

As for official conversions, AFAIK there's only the SEMIC work which converts from iso19139 to DCAT-AP right?

josegar74:
I hope I'm not adding too much noise, but would it be an idea to support xlst-based formatters and formatters based on Elasticsearch index properties (pivot format)? 

Maybe we could start with the current xslt-based ones, keeping open the possibility of supporting formatters based on Elasticsearch index properties. For example, formatters that produce JSON would probably be easier to implement in Java using some JSON library and using the index information, rather than using xslt.

jahow:
@josegar74 both can be done but I really think that having just one formatting system to maintain would be easier in the long run. I have the impression like over the years GN4 has moved from XSLT to index-based formatting, and I think it would make sense to embrace this new approach in GN5 once and for all.

jahow:
> Would it be feasible, in your opinion, to migrate the formatters as-is for now, and then tackle the proposed changes in a follow-up PR? Or do you see benefits in addressing everything at once?

In my opinion yes, although it wouldn't follow the agreed principle of doing proposals for major features in GN5, which in turn shows a lack of consideration with the GeoNetwork community.

PascalLike:
> In my opinion yes, although it wouldn't follow the agreed principle of doing proposals for major features in GN5, which in turn shows a lack of consideration with the GeoNetwork community.

Thanks @jahow But I had the opposite impression. I think the fact that this discussion has been going on for months is exactly the sign that in the Geonetwork community you give the right space to discuss things! 
I understand your concerns about the complexity of supporting both approaches as Jose suggested, but it looks like for now both alternatives are needed. Jose's idea seems as a way to implement your approach progressively, rather than all at once.

fxprunayre:
About this PR, the goal is a "Draft evaluation on how to migrate the formatter API." based on questions raised in https://github.com/geonetwork/geonetwork/issues/66 to try to get a better idea about how to improve that module.

I would say that there is some kind of misunderstanding around GN5 and definitely a lack of discussion around the goal. So far in the last 6 months, the idea was to try to clarify how such a migration can be done in this repository which is "Architecture experiment for GeoNetwork 5 development setting up clean Spring Boot application for Java 21 development". Some of us have been investigating technical challenges to try to get a better picture of how to tackle all those challenges (maybe too many).

It would probably require discussions about the list of modules/features described https://github.com/geonetwork/geonetwork/issues with a draft effort of describing the current status, raising questions and actions. Then maybe the main question would get some kind of answer: What is GN5? A more modularized GN4 to a new technical base or something else. The initial assement was GN4 has CVE or EOL libraries - how to overcome that.

For now, this repository should not be taken for granted, I think, it is still a place where we've been learning/testing/experimenting and there is no certitude even if some aspects are getting clearer (eg. OGC API Records, Security).

jahow:
@fxprunayre are you saying that this repository is not intended to be deployed in production at some point?

In my understanding, GN5 is the future of GN4 and is a rewrite of GN4 to overcome longstanding technical issues.

@ticheler told my colleagues that GN5 would be ready at the end of the year. It didn't sound like it was just an experiment.

fxprunayre:
I'm only saying that there is a number of topic to discuss https://github.com/geonetwork/geonetwork/issues

PascalLike:
The goal of moving the GN code base from the current stack to a Spring boot based one is a delicate and complex task. Making existing things do not break is already a huge technical challenge. I think we should go step by step. Introducing changes where possible is certainly an opportunity not to be wasted, but we must do it without complicating the main operation too much, especially in this first phase.
Having this discussion here must be cumbersome for the developers working on it. I don't even find it nice to have @jahow as the only person defending an alternative point of view, she has valid points that maybe we should discuss in a more holistic way.

Can I create a thread on Gn's Discourse to follow up this discussion? Do you think it could be useful?

jodygarnett:
> Can I create a thread on Gn's Discourse to follow up this discussion? Do you think it could be useful?

Yes that would be perfect, thanks :)

PascalLike:
GN5 Meeting Update: 
What is the status of this? Is it ready for review?

src/shared/security/src/main/java/org/geonetwork/security/AuthenticationFacade.java
```diff
@@ -5,13 +5,20 @@
...
 import org.springframework.security.core.Authentication;
 import org.springframework.security.core.context.SecurityContextHolder;
+import org.springframework.security.core.userdetails.User;
 import org.springframework.stereotype.Component;
 
+/** implementation of IAuthenticationFacade */
 @Component
 public class AuthenticationFacade implements IAuthenticationFacade {
 
+    @Autowired
```
PascalLike, line 19: A best practice is to avoid field injection in favor of constructor injection

PascalLike:
It compiles without any issue, but when I run it fails with this exception:

```
java.lang.IllegalStateException: Failed to introspect Class [org.geonetwork.config.WebSecurityConfiguration] from ClassLoader [org.springframework.boot.loader.launch.LaunchedClassLoader@6ce253f1]
...
Caused by: java.lang.NoClassDefFoundError: UserRepository
```

PascalLike (approved):

davidblasby:
I don't think this is ready to merge.
At least the security stuff needs working on.

PascalLike:
> I don't think this is ready to merge. At least the security stuff needs working on.

@davidblasby please add more details (or link relevant issues).

jodygarnett:
meeting update: this is the oldest PR, in draft to prevent merge.

David indicates all this work is in the code sprint one, can this be removed?

PascalLike:
Merged https://github.com/geonetwork/geonetwork/pull/97

PR #80, opened 2024-12-24
Add data directory and metadata manager components

josegar74:
* Evaluate how to migrate GN4 API on attachment and batch edit API
* Add metadata manager 
* Add data directory

image


src/shared/data-directory/src/main/java/org/geonetwork/metadatastore/MetadataResource.java
```diff
@@ -0,0 +1,33 @@
...
+
+    String getId();
+
+    String getUrl();
+
+    MetadataResourceVisibility getVisibility();
+
+    long getSize();
+
+    Date getLastModification();
```
ianwallen, line 20: It would be nice if all api's in GN5 would use iso dates.  In GN4, this date is shown as a unix integer and it causes issues when using codegen and is also not human readable.

I have been wanting to change this in GN4 however I don't know how to change it without breaking backward compatibility.

juanluisrp:
Spotbugs is preventing the build to finish with this warning:
```
Error:  Medium: Exception thrown in class org.geonetwork.utility.date.ISODate at new org.geonetwork.utility.date.ISODate(String) will leave the constructor. The object under construction remains partially initialized and may be vulnerable to Finalizer attacks. [org.geonetwork.utility.date.ISODate, org.geonetwork.utility.date.ISODate] At ISODate.java:[line 82]At ISODate.java:[line 82] CT_CONSTRUCTOR_THROW
Error:  Failed to execute goal com.github.spotbugs:spotbugs-maven-plugin:4.8.6.2:check (spotbugs-check) on project gn-utility: failed with 1 bugs and 0 errors -> [Help 1]
```

fxprunayre (approved):

PR #82, opened 2025-02-10
GN5 Proxy to GN4: proxy filter predicate for GN4 portals

davidblasby:
@fxprunayre was looking at changing some of the location of the GN5-proxy-to-GN4 (`cloud: gateway:  mvc` in `application.yml`).  One of the issues was dealing with the different GN4 portals (i.e. GN DB table `sources`).

This is a simple spring filter predicate that is aware of the defined Portals (subportals etc...).

```
  cloud:
    gateway:
      mvc:
         routes:
           - id: geonetwork_route
             uri: ${geonetwork.core.url}
             predicates:
               - isGnPortalRequest=/geonetwork
             filters:
               - addSimpleGn4SecurityHeader=gn5.to.gn4.trusted.json.auth
```

The interesting part is:

```
      predicates:
               - isGnPortalRequest=/geonetwork
```

The new predicate is `isGnPortalRequest` and the argument is the context path ("/geonetwork").
It will proxy requests like `"/geonetwork/srv/..."` and `"/geonetwork/<portal name>/..."` where `<portal name>` is the UUID of a `source` in the `sources` db table.

The underlying issue is that there are link the HTML created by GN4 that will not be correct if its in the "wrong" context path.  This helps with this problem, but doesn't fully solve it.

### DIFF
--- a/src/apps/geonetwork/src/main/java/org/geonetwork/gn4proxyprediates/GnPortalPredicate.java
+++ b/src/apps/geonetwork/src/main/java/org/geonetwork/gn4proxyprediates/GnPortalPredicate.java
@@ -1,0 +1,79 @@
+/*
+ * (c) 2003 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license,
+ * available at the root application directory.
+ */
+package org.geonetwork.gn4proxyprediates;
+
+import io.micrometer.common.util.StringUtils;
+import java.util.List;
+import org.geonetwork.domain.repository.SourceRepository;
+import org.geonetwork.utility.ApplicationContextProvider;
+import org.springframework.cloud.gateway.server.mvc.common.Shortcut;
+import org.springframework.context.ApplicationContext;
+import org.springframework.web.servlet.function.RequestPredicate;
+
+/**
+ * Looks at the incoming request and sees if it looks like a GN4 request to one of the GN portals (as defined in
+ * DB:`sources` table). Please see the uuid field in the `sources` table.
+ *
+ * <p>routes: - id: geonetwork_route uri: ${geonetwork.core.url} predicates: - isGnPortalRequest=/geonetwork filters: -
+ * addSimpleGn4SecurityHeader=gn5.to.gn4.trusted.json.auth
+ *
+ * <p>This will match `/geonetwork/srv/...` and `/geonetwork/srv` (default portal name) If you have a portal with ID
+ * (uuid) 'dave', then it will also match `/geonetwork/dave/...` and `/geonetwork/dave`.
+ *
+ * <p>If there is no source called "home", then `/geonetwork/home/...` will NOT MATCH.
+ */
+public class GnPortalPredicate {
+
+    static ApplicationContext applicationContext = ApplicationContextProvider.getApplicationContext();
+
+    /*
+     * This does the matching to things like '/geonetwork/<portal uuid>/...'.
+     * This also matches the default portal '/geonetwork/srv'.
+     *
+     * @param contextPath - typically "/geonetwork"
+     * @return true if request path is like "/geonetwork/PORTAL-UUID/..."
+     */
+    @Shortcut
+    public static RequestPredicate isGnPortalRequest(String contextPath) {
+        return request -> {
+            if (!request.path().startsWith(contextPath + "/")) {
+                return false; // not to "/geonetwork/..."
+            }
+            // remove "/geonetwork" from start
+            var path = request.path().substring(contextPath.length());
+            if (path.startsWith("/")) {
+                path = path.substring(1);
+            }
+            if (StringUtils.isBlank(path)) {
+                return false;
+            }
+            // portalName: "/geonetwork/srv/..." --> "srv"
+            var portalName = path;
+            if (portalName.contains("/")) {
+                portalName = portalName.substring(0, portalName.indexOf("/"));
+            }
+            if (portalName.equals("srv")) {
+                return true; // this is the default portal
+            }
+            var portalIds = portalIds();
+            return portalIds.contains(portalName);
+        };
+    }
+
+    /**
+     * Get the IDs (UUID) of all the portals (DB: `sources` tables).
+     *
+     * <p>Note, in general, the UUID of the main portal will be GUID.
+     *
+     * @return the IDs (UUID) of the all portal (DB: `sources` tables)
+     */
+    static List<String> portalIds() {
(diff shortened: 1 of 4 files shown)


davidblasby:
@fxprunayre - as we discussed monday evening.

fxprunayre:
It works but did you try https://www.baeldung.com/spring-cloud-gateway-routing-predicate-factories ?
to make it non static (and to be able to cache the Source repo query which currently is executed on each request)

davidblasby:
I did do something similar for the other proxy filters.  I will take a look and see if I can cache the repo to speed it up (although I doubt it actually takes much time since the DB will have cached the table).

fxprunayre:
> if I can cache the repo to speed it up (although I doubt it actually takes much time since the DB will have cached the table).

No idea what is the cost of it (with a source table with 230 rows), but I was having lot of SQL in the log for this 
```
2025-02-13T20:13:00.565+01:00 DEBUG 651974 --- [GeoNetwork] [nio-7979-exec-6] org.hibernate.SQL                        : select s1_0.uuid,s1_0.creationdate,s1_0.filter,s1_0.groupowner,s1_0.islistableinheaderselector,s1_0.logo,s1_0.name,s1_0.servicerecord,s1_0.type,s1_0.uiconfig from sources s1_0
2025-02-13T20:13:00.565+01:00 DEBUG 651974 --- [GeoNetwork] [io-7979-exec-10] org.hibernate.SQL                        : select s1_0.uuid,s1_0.creationdate,s1_0.filter,s1_0.groupowner,s1_0.islistableinheaderselector,s1_0.logo,s1_0.name,s1_0.servicerecord,s1_0.type,s1_0.uiconfig from sources s1_0
2025-02-13T20:13:00.565+01:00 DEBUG 651974 --- [GeoNetwork] [nio-7979-exec-7] org.hibernate.SQL                        : select s1_0.uuid,s1_0.creationdate,s1_0.filter,s1_0.groupowner,s1_0.islistableinheaderselector,s1_0.logo,s1_0.name,s1_0.servicerecord,s1_0.type,s1_0.uiconfig from sources s1_0
```

Some caching was tested in https://github.com/geonetwork/geonetwork/blob/main/src/shared/utility/src/main/java/org/geonetwork/utility/schemas/CodeListTranslator.java#L32

davidblasby:
@fxprunayre  - I tried really really really really really hard to use the `AbstractRoutePredicateFactory`, but I am unable to get the predicate factories to register with spring for use in `application.yml`.  I tried to get the example you linked to work (and 2 others), but they are not registering either.  If we want to go with those, I think we'll have to stop using the route configs inside `application.yml`, and config in java.  That's much better supported.

I can, however, add support for caching to the current infrastructure.

davidblasby:
@fxprunayre - I added a cache.  Its set for 60 second timeout.  
Postgresql will typically execute a query in about 10ms if the table is cached in the OS - but this will cut down on traffic to and from the DB.

davidblasby:
I used guava's cache - however, the `@Cachable` might work too.  Not sure how multiple caches handle each other and how to set the eviction info for them.

fxprunayre:
Some suggestions in https://github.com/davidblasby/geonetwork/pull/2 to cleanup routing between GN5&4. I have some difficulties with static method which works in test and not runtime - maybe you've ideas to improve that?

Using same context path for both will facilitate transition probably. 

About caching, we should probably choose one option to avoid what we have in GN4 (JCS+Guava+Spring cache).

Also tried the AbstractRoutePredicateFactory and it is probably due to the fact that it is for reactive server and not mvc.

davidblasby:
@fxprunayre - see comments in #2.  I will look into this next week.

wrt `AbstractRoutePredicateFactory` - that would make sense if its not for MVC.   Too bad...

davidblasby:
@fxprunayre  - I've merged in your changes + added some more.

1. Its using Spring Caching
2. I've removed some of the Static stuff (no longer need the ST_WRITE_TO_STATIC_FROM_INSTANCE_METHOD suppress).  However, the spring filter stuff doesn't make it easy because its based on static methods.
3. I added a cache test

davidblasby:
Clean up the spring config in the test cases so its easier to understand.

davidblasby:
resolved merge conflicts

config/application.yml
```diff
@@ -11,6 +11,17 @@
 #
 
 geonetwork:
+  url: http://localhost:7979
```
PascalLike, line 14: There is a duplicate **url** value on line 75 of the file. Please check

src/apps/geonetwork/src/main/java/org/geonetwork/gn4proxypredicates/CachingPortalIds.java
```diff
@@ -0,0 +1,58 @@
...
+ *
+ * <p>See the CacheLoader.load() function. This actually makes the request to the DB to get the list of portal UUIDs.
+ *
+ * <p>The cache is set to expire this after CACHE_TIMEOUT_SECONDS seconds. After this time, it will re-execute the db
+ * request for the portal uuids.
+ */
+@Component
+public class CachingPortalIds {
+
+    @Autowired
```
josegar74, line 27:
```suggestion
```

I don't think `@Autowired` is required here.

src/apps/geonetwork/src/main/java/org/geonetwork/gn4proxypredicates/CachingPortalIds.java
```diff
@@ -0,0 +1,58 @@
...
+    SourceRepository sourceRepository;
+
+    @Autowired
```
josegar74, line 30:
```suggestion
```

I don't think `@Autowired` is required here.

src/apps/geonetwork/src/main/java/org/geonetwork/gn4proxypredicates/CachingPortalIds.java
```diff
@@ -0,0 +1,58 @@
...
+        return portalUUIDs;
+    }
+
+    //    @Autowired
+    //    CacheManager cacheManager;
+    //
+    //    /** Clean cache NOW. */
+    //    public void clear() {
+    //        cacheManager.getCache("source-portal-ids").invalidate();
+    //    }
```
josegar74, line 53:
```suggestion
```

Remove unused code

davidblasby:
ok - fixed the little issues here.  Nothing major changed.

PascalLike:
Meeting update: will be reviewed soon and eventually merged

PR #83, opened 2025-02-19
Handle XSRF token through a fetch middleware

jahow:
Instead of having hardcoded XSRF tokens when calling the GN4 API.

I haven't encountered the issue with the GN5 API but it might be worth it to do the same.

Note that this is required because the code generated using OpenAPI is not Angular code; otherwise the XSRF token handling would be done automatically by Angular.

### DIFF
--- a/geonetwork-frontend/projects/glib/src/lib/config/config.loader.ts
+++ b/geonetwork-frontend/projects/glib/src/lib/config/config.loader.ts
@@ -1,6 +1,7 @@
 import {
   Configuration,
   DefaultConfig,
+  Middleware,
   SiteApi,
   UiApi,
   UiConfiguration,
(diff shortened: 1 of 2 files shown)


jahow:
@fxprunayre @juanluisrp not sure why the tests are failing, I hope this fix makes sense to you! I couldn't get any API calls to work otherwise

fxprunayre (approved):
Thanks for this, @jahow! 

> code generated using OpenAPI is not Angular code;

Yes, it was one of the questions about the openAPI generator, in one place we need to load the app config before Angular initialization, but then indeed the API will be used in the context of Angular. So maybe make more sense to generate an Angular client. It probably brings some benefits...

jahow:
> Yes, it was one of the questions about the openAPI generator, in one place we need to load the app config before Angular initialization, but then indeed the API will be used in the context of Angular. So maybe make more sense to generate an Angular client. It probably brings some benefits...

I would recommend this yes, although that will need some refactoring of the code. Otherwise the app will end up using a mixture of XHR and Fetch wrapped by different constructs. The XSRF token was a good example of that, although admittedly it was a straightforward fix.

PR #84, opened 2025-02-19
better oidc/oauth2 login support

davidblasby:
This improves the OIDC/OAuth2 support for GN5.

1. I've moved the create-the-user-in-the-gn-db from the attribute mapper to its own service (spring's `DefaultOIDCUserServer` and `UserService` (for OAUTH2)).  This service (`GeoNetworkOAuth2UserService`), basically, does two things - (a) creates the user in the DB if its not already there (b) creates the permissions (authorities) from the DB.

2. I didn't like how GitHub was defaulting to a number for a username, so I've put in documentation how to change it (to `login` attribute in the token).

3. I removed the old `userOauthAuthoritiesMapper` - the work is now done with `GeoNetworkOAuth2UserService`.  The mapper needed information that wasn't readily available to it.  I think the mapper is supposed to map existing permissions, not pull them from the database.

4. Updated the header-appender (for the proxy filter) so its aware of OIDC and OAUTH2 authentication types.

5. I was originally using the `SPRING_SECURITY_CONTEXT ` Session variable (provided by SPRING) for the current authentication token.  This didn't work for OIDC, so I've moved it to use the `SecurityContextHolder`.  There's a possibility that this will cause an issue, but I've tested and havent noticed anything.

6. I added support for customizing the way the OIDC/OAuth2 token's attributes are used to construct the GN DB User.  The registration name's should match the one you use in the   `security: oauth2: client: registration:` configuration.

```
geonetwork:
  sso:
    registration:
      github:
        email: email
        name: login
        organization: company
        surname: lastname
      custom:
        email: email
        name: given_name
        organization: azp
        surname: family_name
```

7. I changed the `authType` to null.  See GN4:

https://github.com/geonetwork/core-geonetwork/blob/main/core/src/main/java/org/fao/geonet/kernel/security/GeonetworkAuthenticationProvider.java#L87

NOTE: this changes how the usernames are extracted and are stored in the database.  This shouldnt be a problem, however, existing users might be an issue (depending on how your IDP is configured).


src/apps/geonetwork/src/main/java/org/geonetwork/config/GeoNetworkSsoConfiguration.java
```diff
@@ -0,0 +1,44 @@
...
+import java.util.HashMap;
+import java.util.Map;
+import lombok.AllArgsConstructor;
+import lombok.Getter;
+import lombok.Setter;
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration
+@ConfigurationProperties(prefix = "geonetwork.sso")
```
fxprunayre, line 17: Should we move this to geonetwork.security.oauth2 to make it relates to spring.security.oauth2 ? cf. https://github.com/geonetwork/geonetwork/blob/main/config/application.yml#L30

davidblasby: ok, moved to this, and I changed the name of the config item so it was a bit more clear.

```
geonetwork:
  security:
    oauth2:
      registration:
        github:
          email: email
          name: login
          organization: company
          surname: lastname
        custom:
          email: email
          name: first_name
          organization: company
          surname: last_name
```

Also, fixed issue with the anonymous user.

fxprunayre (approved):
LGTM. Works fine with https://aai.egi.eu/

PR #85, opened 2025-03-03
Formatter security and access manager

davidblasby:
Hi, @fgravin ,

This is a PR-to-your-PR.

1. brought this up-to-date with the `main` branch  (which also has some security stuff in it)
2. please see the notes in the docs (will integrate when we merge the docs stuff from OGCAPI-Records).

Basically, this:
1. More strict useage of the a standard security system (`authenticationFacade`)
2. All security types put the permissions in the `SecurityContext` and are built from the db (`user` and `usergroup` tables)
3. Don't use `ROLE_*` like permissions
4. Most code shouldn't use `UserRepository` and definitely shouldn't be using `UsergroupRepository`
5. Added more functionality to the `MetadataAccessManager` (works similar to GN4)


davidblasby:
@fxprunayre  - I think this is ok to merge into your `FormatterApi` branch.

PR #92, opened 2025-05-07
Build / Fix warning due to duplicated gn-index-model dependency ref

fxprunayre:
```
Some problems were encountered while building the effective model for org.geonetwork:gn-ogcapi-records-openapi-autogen:jar:5.0.0-SNAPSHOT
[WARNING] 'dependencies.dependency.(groupId:artifactId:type:classifier)' must be unique: org.geonetwork:gn-index-model:jar -> duplicate declaration of version ${project.version} @ line 85, column 17
```


davidblasby:
missed this - thanks for catching it!

PascalLike (approved):
Thanks Francois

PR #94, opened 2025-05-30
API / Formatter migration improvements

josegar74:
DO NOT MERGE

This is incorporated into - https://github.com/geonetwork/geonetwork/pull/96

- End-point to return available formatters for a metadata.
- Update DCAT formatter and add DCAT-AP formatters.
- Add DCAT / DCAT-AP formatters for ISO19139.
- Support for formatters based on IndexRecord instead of XML / XSLT
- Update the formatter XSLT to version 3.0.
- Initial version of SettingManager.


pom.xml
```diff
@@ -213,6 +213,11 @@
         <artifactId>spotbugs-annotations</artifactId>
         <version>4.8.4</version>
       </dependency>
+      <dependency>
+        <groupId>com.neovisionaries</groupId>
+        <artifactId>nv-i18n</artifactId>
```
PascalLike, line 218: This is currently [unmaintained](https://github.com/TakahikoKawasaki/nv-i18n/issues/96). Since it's used only in few places, maybe later we can replace it with https://github.com/unicode-org/icu and https://github.com/unicode-org/icu/blob/main/icu4j/main/core/src/main/java/com/ibm/icu/util/ULocale.java ?

PascalLike:
Merged https://github.com/geonetwork/geonetwork/pull/97

PR #95, opened 2025-06-02
Update database model to use boolean converter

josegar74:
# Update database model to use boolean converter

## Description

The current model classes with boolean properties were using `String` fields instead of `Boolean`. This change request uses `Boolean` types (same as in GeoNetwork 4), using an  hibernate converter to store and retrieve the information from a character type, as some databases doesn't support booleans.

## Type of Change
Select one:
- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [X] Refactoring

## Approach

Properties have been changed to `Boolean`, including the related code using them. An hibernate converter has been created to map between the Java type and the database column type: `y` <-> `true`, `n` <-> `false`

## Checklist
- [X] Tests updated (if applicable)

### DIFF
--- a/src/shared/domain/src/main/java/org/geonetwork/domain/converter/BooleanToYN.java
+++ b/src/shared/domain/src/main/java/org/geonetwork/domain/converter/BooleanToYN.java
@@ -7,6 +7,9 @@ package org.geonetwork.domain.converter;
 
 import org.hibernate.type.CharBooleanConverter;
 
+/**
+ * Handles conversion to/from {@code Boolean} as {@code 'y'} or {@code 'n'}
+ */
 public class BooleanToYN extends CharBooleanConverter {
     public static final BooleanToYN INSTANCE = new BooleanToYN();
     private static final String[] VALUES = new String[] {"n", "y"};


src/shared/domain/src/main/java/org/geonetwork/domain/converter/BooleanToYN.java
```diff
@@ -0,0 +1,39 @@
+/*
+ * (c) 2003 Open Source Geospatial Foundation - all rights reserved
+ * This code is licensed under the GPL 2.0 license,
+ * available at the root application directory.
+ */
+package org.geonetwork.domain.converter;
+
+import org.hibernate.type.CharBooleanConverter;
+
+public class BooleanToYN extends CharBooleanConverter {
```
PascalLike, line 10: Even though the code is very clear and self-explanatory, I would suggest to start using Javadoc or comments on new classes, to make it easier for new developers to onboard.

PascalLike (approved):
Thanks Jose, it looks good to me.

src/shared/domain/src/main/java/org/geonetwork/domain/converter/BooleanToYN.java
josegar74: @PascalLike thanks for the review, I have add the Javadoc, please check and merge it if all looks fine.

PR #96, opened 2025-06-02
FormatterApi

davidblasby:
DO NOT MERGE

This replaces https://github.com/geonetwork/geonetwork/pull/78 and include's Jose's changes https://github.com/geonetwork/geonetwork/pull/94.

Based on https://github.com/geonetwork/geonetwork/pull/78:

1. I merged Jose's changes (https://github.com/geonetwork/geonetwork/pull/94)
2. I created new shared module for formatting
3. I moved all the non-http code new shared formatting library
    + Created new FormatterApi which is a high-level equivalent to the controller's api
    + Changed FormatterController so it calls the new API (just handles HTTP)
    + module/formatter now is just the controller

TODO:

1. Add tests:
    + this will be the plan for how all the other modules do some testing
    + I am not sure how to start a test case up with all the schema's accessible, etc...
    + @josegar74 @fxprunayre if you setup a few tests I can likely add some more.
    + would be great to have a test case for each formatter type
2. Add the OGCAPI-Records Elastic-index-to-JSON formatter (convert so it registered)
    + david to do (different PR)
4. Allow formatterApi usage in  OGCAPI-Records
   + david to do (different PR)
5. Look at the new security stuff and validate it as well as move to the security shared module
   + prob best if @josegar74 and @fxprunayre take a look 
7. Not sure if there is proper security checking happening for requesting a formatter record - perhaps we should merge this and worry about it later (@PascalLike can create ticket)

## Type of Change
Select one:
- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [x] Refactoring

## Related Issue
Closes/updates # (optional):

## How to Verify

1. 
5. 
8. 

## Checklist


davidblasby:
I am just about to push a commit for creating the `OgcApiRecordsJson` output - it was really simple (I just wrote a tiny wrapper class around the existing` OgcApiRecordsGeoJsonConverter`).  It appears to be mostly working.

However, one thing I noticed has to do with the response MIME Type.

This seems to be generated from the `FormatType` enum (see ` FormatType#findByFormatterKey()`:

```
public enum FormatType {
    txt("text/plain"),
    html("text/html"),
    xml("application/xml"),
    json("application/json"),
    jsonld("application/vnd.schemaorg.ld+json"),
    pdf("application/pdf"),
    testpdf("application/test-pdf");
```

The problem is that the ContentType is disconnected from the code that's doing the actual conversion. 

For example, my `OgcApiRecordsJson` converter creates JSON (`application/json`) output.  I think this should be part of the `IndexFormatterProcessor` interface (`#getMimeType()`) as well as the other interfaces.  Otherwise, to add a new output format, I'd have to modify `FormatType`.

davidblasby:
I pushed a commit to:

1. add the `FormatterController` (2 endpoints) to the main GeoNetwork App 
    + this is just adding a dependency to the `app/geonetwork` app
2. Add the ogcapi-records specified GeoJSON output as an Index-based Output 
    + although this is an-index based output, i think it still puts the whole Metadata XML in the output (need to check)

fxprunayre:
> * this is mostly to see how to properly setup a test case that has all the schema info in it.

https://github.com/geonetwork/geonetwork/tree/features/frontend-exp-indexrecord/src/modules/data-analysis/src/test/java/org/geonetwork/data/gdal contains some examples of tests to evaluate how to wire required dependencies depending on the type of test.

> Where to get Formatter's Mime Type (See above). My GeoJSON output doesn't set the correct mime type (and I'm not sure if its easy)

In microservices work, we made a link between a formatter and a mimetype which is maybe something to have in GN5
https://github.com/geonetwork/geonetwork-microservices/blob/fc14592921e6759aa6cf702a07e9394a3704cbff/modules/library/common-search/src/main/resources/application.yml#L88-L89

Maybe the only exception would be PDF which could be used on various formatters but that's a specific story probably.

In GN4, we also have some `config.properties` file in formatter folders that was used to store additional formatter info
https://github.com/geonetwork/core-geonetwork/blob/main/schemas/iso19139/src/main/plugin/iso19139/formatter/

PR #109, opened 2025-09-01
Adding JSON-LD support for records and collections

PascalLike:
# Adding JSON-LD support for records and collections

## Description
This PR introduces an additional format for representing metadata in JSON-LD format. The format mirrors the DCAT-AP 3.0 fields and is validatable using basic validators (https://www.itb.ec.europa.eu/shacl/dcat-ap/upload).

This format is available via OGC API Records, as an alternate link for all metadata.

## Type of Change
Select one:
- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Refactoring

## Related Issue
Closes/updates # (optional):

## Approach
The new format is generated from the data in the IndexRecord, so it is based on the index and not the native metadata format. This makes the formatter available for any indexed metadata, regardless of the source schema.

## How to Verify

1. 
2. 
3. 

## Checklist


davidblasby (approved):
This looks interesting - I think the code is fine.

I'd like to see @josegar74 look at it from a DCAT perspective.  Does this handle the different dcat types?

PascalLike:
No longer applicable

PR #110, opened 2025-09-02
Remove unnecessary parameters from the metadata formatter controller

josegar74:
# Remove unnecessary parameters from the metadata formatter controller

## Description

The formatter controller used a parameter `profile` that did not make sense anymore.At one point, the possibility of having formatters with the same ‘name’ (formatterId) and different profiles, to be used by the OGC API's log API, was discussed. However, each formatter name/profile is unique, so it does not make sense for the formatter controller to accept both.

The ‘mdpath’ parameter was also removed, as it was not used and its potential use was unclear.

The ‘iso3lang’ parameter was retained, as it will likely be necessary to support multilingua.

## Type of Change
Select one:
- [] Bug fix
- [ ] New feature
- [ ] Documentation update
- [X] Refactoring

## Checklist


davidblasby (approved):
I think this is okay - I don't think it will have any impact on ogcapi-records.

Is this going to have any issues for the GN4->GN5 work?

PR #111, opened 2025-09-03
API / Search / User privilege filter

fxprunayre:
## Description

* Refactor to reuse OGC API Records user privileges search filter in main search API
* Add documentation for search endpoints

image

## Type of Change
Select one:
- [x] Bug fix
- [x] New feature
- [x] Documentation update
- [x] Refactoring

## Related Issue
Closes/updates # (optional):

## How to Verify

1. 
2. 
3. 

## Checklist


davidblasby:
My goal would be to remove this search api and use the ogcapi-records one (which is GREAT with queryables, CQL, facets, and sortby - and dynamic properties).

However, I expect this search api to be used by the GN4->GN5 work.

fxprunayre:
If users want to also keep the capability to have third party tools to easily connect to GN index (eg. kibana, superset), it is good keep an Elasticsearch endpoint which provides much more flexibilty than OGC API Records. And indeed during GN4>5 transition, this is needed also.

PR #120, opened 2025-10-15
Modularize

fxprunayre:
# Modularize

## Description

Moving main app content into dedicated modules in order to be able to compose new apps more easily.

Actions:
- [x] Move proxy to a module
- [x] Move GN4 forwarding and security token to a module
- [x] Move configurations in a common module and create a BaseApplication
- [x] Create an OGC API Records app only
- [ ] How to load only application*.yaml required for an app? (eg. OGC API Records does not need the GN cloud proxy config)

## Type of Change
Select one:
- [ ] Bug fix
- [ ] New feature
- [ ] Documentation update
- [x] Refactoring

## Related Issue
Closes/updates # (optional):

## How to Verify

1. 
2. 
3. 

## Checklist


fxprunayre:
Good. About loading some of the `application*.yaml` in the various apps, maybe we should use profiles instead of the current import, and enable them on a per app basis? I'm not really sure how to do that.

davidblasby:
> Good. About loading some of the `application*.yaml` in the various apps, maybe we should use profiles instead of the current import, and enable them on a per app basis? I'm not really sure how to do that.

Yes - the `application.yaml` stuff should be configurable from the various modules.  Thats on my list of things to look into.

davidblasby:
@fxprunayre @PascalLike - I'm a bit distracted today, and have been thinking about this.

I think we are moving in the correct direction.

However, I think we should:

1. have a `/apps/generic-geonetwork5` that will have basic support in it for the gn5 app.  That means the support for the endpoints - things like a general `application.yaml` and support being able to bring in multiple `application.yaml` configurations from the sub-module (i.e. ogcapi), security, content negotiation, etc..  In general, this will not do too much, but will be the system that the sub-modules plug into.

2. Other apps (like GN4-replacement, ogcapi-records, csw, FAIR validation, vocab management, etc...) would basically just present class (plugins) that the  `generic-geonetwork5` would pickup and make sure the system is configured correctly.  I would think these would live in the `/apps/` directory.  These wouldn't be large as they would just be supporting the spring-boot application.

3. `/modules/...` as we have now - these would be mostly be the controller & endpoint DEFINITIONS.  For example, the ogacapi-records module is, basically, just the auto-generated swagger application (including controller and DTO definitions).

4. `/shared/...` as we have now - these are sharable implementations so others can re-use code blocks.  95% of the ogcapi-records code is here.

I think the main difference between this and what @fxprunayre is suggesting is a separation between `module` and `app`.

Just to be clear, for ogcapi-records,

1. main spring-boot application source code will be in  `generic-geonetwork5`
2. in `/apps/ogcapi-records` will be class that will plug into `generic-geonetwork5` with specific ogcapi-records config.  If you look at the `main` branch, you'll see there is code in `/apps/geonetwork` configuring some things for the ogcapi-records app.  In the "new" way, this will live in  `/apps/ogcapi-records`
3. `/modules/ogcapi-records` (as now) and `/shared/ogcapi-records` (as now)

Theoretically, it would be possible to create a new app (that's ogcapi-records and csw) just by creating a new module that brought in the `ogcapi-records` and `csw` module and everything would just plugin to the `generic-geonetwork5` app.  

Another example would be if you want `ogcapi-records` and `proxy-ed gn4` - you'd create a new app and just have the pom.xml bring in ogcapi-records (app) and proxy-ed-gn4 (app).

You could even have an app that configured OIDC single-sign-on and then just bring that config into the system.

This, should, allow for a lot of reuse and configuration ease.

However, we'll have to see how to do proper isolation (i.e. so ogcapi-record's jackson `ObjectMapper` doesn't clash with another module's `ObjectMapper` configuration) and combination of multiple `application.yaml`.

Let me know what you think - I think this is on track with what we've been discussing.

PR #145, opened 2026-01-29
Fix for the XmlMapperConfig bean creation at startup

tanzeeladebont:
Suggested fix for startup error . 

```
***************************
APPLICATION FAILED TO START
***************************

Description:

Field xmlMapper in org.geonetwork.ogcapi.ctrlreturntypes.OgcApiRecordsCollectionsResponseFormatter required a bean of type 'com.fasterxml.jackson.dataformat.xml.XmlMapper' that could not be found.

The injection point has the following annotations:
 ```


PascalLike:
Thanks Tanzeela, this could work as a temporary patch. I think it's better to understand why this class is missing (or why it's required).

PascalLike:
@tanzeeladebont please check this https://github.com/geonetwork/geonetwork/pull/146

tanzeeladebont:
Thank you @PascalLike . I will close this PR. Thank you for providing  the right fix.

Issue #87, opened 2025-04-29
How to document GeoNetwork 5?


PascalLike:
Developer documentation is currently located in the directory:

https://github.com/geonetwork/geonetwork/tree/main/docs

Developer documentation should contain:
- How to setup a workspace
- Technical instructions on the various ways to configure Geonetwork and integration with other tools
- Documentation of common modules and APIs.

We need to decide whether to continue with current approach and give some indications for future writers (format, level of detail etc.).
Good developer documentation must be aligned to [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) principle.

## Actions

- [ ] Decide where to write developer docs
- [ ] Write a guideline for developer docs

PR #157, opened 2026-03-30
Feature/developer documentation 1

PascalLike:
# Developer documentation

## Description
This introduces some changes to documentation to outline the system architecture of GeoNetwork 5 backend. Its primary objectives are to:

- Provide a comprehensive overview of the platform's core features, underlying technology stack, and design patterns.
- Onboard Technical Staff.

## Type of Change
Select one:
- [ ] Bug fix
- [ ] New feature
- [X] Documentation update
- [ ] Refactoring

## Related Issue
https://github.com/geonetwork/geonetwork/issues/87

## Approach
The idea is to integrate future changes/decisions in this documentation with centralized [Architecture Decision Record (ADR)](https://martinfowler.com/bliki/ArchitectureDecisionRecord.html). 
Every architectural, functional, and systemic design choice made during development will be tracked together with this, to preserve context and rationale (I will itroduce this in another PR)

## Checklist
- [X] Documentation updated (if applicable)


docs/manual/docs/Developer/img/system_1.png
davidblasby, line 1: Might want to remove the "File System" from this image.

docs/manual/docs/Developer/api-guidelines.md
```diff
@@ -0,0 +1,31 @@
+# REST API Design Guidelines
+
+This section defines the routing design principles for the GeoNetwork 5 REST API. Our primary goal is to build on the [OGC API record specification](https://github.com/opengeospatial/OGC-Web-API-Guidelines?tab=readme-ov-file#design-principles), extending it with intuitive, predictable, and highly readable interfaces, without compromising standard compatibility.
+
+
+1. A well-designed URL should allow a developer to locate related endpoints, easily edit paths manually, and share readily understandable links.
+2. No redundant prefixes, technology-exposing file extensions, or unnecessary verbosity.
+3. All new endpoints must either adhere directly to the OGC API structure (*/collections/{collectionId}/items*) or mirror its structural rhythm for custom extensions.
+4. Paths should use simple, clear nouns. Avoid heavy symbol usage or unnecessary hex encoding where possible.
+5. Identity via Path, Not Query Parameters
```
davidblasby, line 10: OGCAPI-Records implementation uses query parameters in two areas:

First is for search (This seems to be the REST-API recommended way to do searches):

There are A LOT of query parameters in ogcapi-records.  Here is a parsed version of the search (query) parameters:

https://github.com/geonetwork/geonetwork/blob/main/src/shared/ogcapi-records/src/main/java/org/geonetwork/ogcapi/service/querybuilder/OgcApiQuery.java

Second, query parameters are also used for output format and profile.  You can, technically, put some/all of this is the header - however, this makes the "link" section more difficult since it would have to specify multiple headers for a request.  The content negotiation system in GN5 is quite complex because it usage is quite different from how most people do output formats (format/profile-different-objects vs format-a-common-object).

Both of these are a cross-concern to the REST (id-based) system.

I think you changed it to "Identify via Path", which is 100% correct!!!

docs/manual/docs/Developer/api-guidelines.md
```diff
@@ -0,0 +1,31 @@
...
+* **GN5 Standard**: */api/collections/maps/items/54896*
+6. Navigate down the resource tree logically. The resource nesting should match the OGC API resource tree, extending it cleanly for custom GeoNetwork features.
+
+    For example:
+
+* **List collections**: */api/collections*
+* **Specific collection**: */api/collections/{collectionId}*
+* **Items in collection**: */api/collections/{collectionId}/items*
+* **Specific item**: */api/collections/{collectionId}/items/{itemId}*
+7. When a resource exists within a user's specific context or multiple relationships, express this cleanly in the path as an extension of the base OGC model. It may be useful, just for example, when a workflow is active and user copies of documents exist.
```
davidblasby, line 21: This doesn't conform to theOGCAPI-Records specification.

The [ogcapi-core spec](https://docs.ogc.org/is/19-072/19-072.html#rc_oas30-security) doesn't talk a lot about security.

```
The OpenAPI specification currently supports the following [security schemes](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#security-scheme-object):

HTTP authentication;
An API key (either as a header or as a query parameter);
OAuth2’s common flows (implicit, password, application and access code) as defined in RFC6749; and
OpenID Connect Discovery.
``` 

The [security schemes](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md#security-scheme-object) site has more details.

Currently, we allow people to login and then control what they see in a request based on their user roles (in the same manner as GN4). 

There isn't any current mechanism to look at another user's collections (which *might* be useful for the `super-user`, but I expect they would use the user admin system for that - as in GN4).

Could you elaborate on this more?  I haven't seen `/users/{username}/` in an api before (except for ACTUAL user admin like in keycloak's API).  Its always been implicit with the permissions of the logged in user.

docs/manual/docs/Developer/api-guidelines.md
```diff
@@ -0,0 +1,31 @@
...
+* **User Spaces**: */api/users/{username}/collections/{collectionId}/items*
+
+
+## Conformance classes
+
+To maintain strict consistency with the OGC API architecture, the advertisement of implemented API features, custom extensions, and supported data schemas must be handled at the [conformance classes](https://docs.ogc.org/is/17-069r3/17-069r3.html#_declaration_of_conformance_classes) level.
+
+Rather than relying on custom endpoints or ad-hoc queries to expose what the API can do, all capabilities should be declared as URIs within the global */conformance* endpoint. This ensures that any client interacting with the GeoNetwork 5 API can systematically discover both standard OGC functionalities and GN5-specific features (such as supported metadata schemas, specific output formats, or custom operations) in a predictable, standard-compliant manner.
+
+For conformance classes that are specific to GeoNetwork (which do not exist in any official standard) it is possible to manage custom ones in an internal namespace.
```
davidblasby, line 31: The profile-names (i.e. which dcat profile you want) are another missing "thing" that could be handled in the same manner.  Would be nice to have some sort of registry for these.  

Perhaps we need to make a `geonetwork.org/conformance` section in the website that could be used to describe anything custom (including profile-name).

docs/manual/docs/Developer/img/system_2.png
davidblasby, line 1: I like this image, however, I am not sure if there will be "mandatory" apps.  For example, someone might just want the CSW in a outside-the-firewall configuration.

I'm not sure I understand the lines coming from the "main mandatory app".  For scalability, these shouldn't be hard links.  I would expect to the see the Editor App depending on the DBs and communicating with the other apps via a message bus.

I would like to see a more detailed version of this which would show how "apps" would be added to the system (including their own GUI and backend components).

Also, might want to remove the "File System".

davidblasby (changes requested):

docs/manual/docs/Developer/img/system_1.png
PascalLike: Renamed as "Storage", to generalize the concept and make it more flexible for a scalable/distributed architecture, let's discuss later if still necessary.

docs/manual/docs/Developer/api-guidelines.md
PascalLike: I think this is compatible with the OGC API philosophy. For me, /collections/{collectionId} and /collections/{collectionId}/items/{itemId} are perfectly what is meant by Identity via Path. Searching for items in a collection (/collections/{collectionId}/items) via query parameters is filtering, so it is perfectly legal to use query parameters. It would be different if we used query parameters on /collections/{collectionId}/items to identify a single item; this would not be compatible.

docs/manual/docs/Developer/api-guidelines.md
jahow: +1 for having conformances on the geonetwork-opensource.org website, so that the URLs actually resolve to something :)

The conformance URIs could also simply redirect to the GN5 API documentation

PascalLike: 100% Agree! We don't have elements/resources to clarify this now, but we can expand on it later once we've taken some concrete steps in this direction.

docs/manual/docs/Developer/api-guidelines.md
PascalLike: For clarity, this document is only intended to highlight API design principles from what we're doing; it's not  "We need to implement this API endpoint". 

The idea isn't to manage resource permissions by including the username in the URL, of course out of discussion. The idea is to manage contextualized informations about entities in the API/system. The example is about the entity user and its relation with records, let's imagine for a reviewer that needs to list draft records per user; probably, there is a standard way to get the same resource with some search on the collection/items endpoint. The idea here is to promote the expressiveness of the API URL, if user example is confusing, I can change that.
Compared to the OGC API, there are some specifications that use a similar mechanism, see OGC API Styles and Map (/styles you get all the styles and /collections/{collectionId}/styles you get the styles that applies to the collectionId).

docs/manual/docs/Developer/img/system_2.png
PascalLike: I see your point. But I thought that having a minimal entity worthy of being called a "GeoNetwork" could help the system design/deployment. Do we really have the need of shipping only CSW? Then where we create/populate the DB? Where do we initialize the system? 

About the lines, I agree that are an oversimplification.

Issue #165, opened 2026-04-23
Use DB to configure Facets & properties


davidblasby:
Currently, the ogcapi-records Facets (and properties) configuration are stored in .yaml files. 

Example:

https://github.com/geonetwork/geonetwork/blob/2908d3bc9d76dd3579272343ba1961a25e4ab45a/config/application-ogcapi-records.yml#L6-L15

This isn't ideal because its static and difficult to modify.

**New feature:**

a) refactor the config for maintaince
b) move the configuration to DB
c) create mechanism for telling modules that the configuration has changed  (could be in other containers or external systems)
     * start with something simple - have update sequence in the database
     * later - message queue
d) Update OpenAPI endpoints (RnD - not sure if they allow changes)
e) REST endpoints

Issue #156, opened 2026-03-24
Configuration management


PascalLike:
### Objectives:

- To have an alternative central way to manage apps configurations like [this](https://github.com/geonetwork/geonetwork/issues/165), without rely on application yamls.
- Allowing to change the configurations at runtime, notifing multiple apps.

### Proposed solution:

Geonetwork 4's configuration was managed in the database using two tables, one with a key-value approach (`SETTINGS`) and the other using structured JSON objects (`SETTINGS_UI`). With the adoption of Spring Boot and a multi APP model, this approach would need to be adapted to avoid technical limitations.

The Spring boot best practice is [Spring Cloud Config Server](https://docs.spring.io/spring-cloud-config/docs/current/reference/html/) to externalize and centrally manage domain and operational configurations. Spring Cloud Config provides a unified server to manage properties for all applications, ensuring consistency and providing an update mechanism to notify other modules/apps.

By default Spring Cloud Config manages configuration on Git, which is, nice but introduces an extra dependency (GIT). Spring Cloud Config also supports [JDBC](https://docs.spring.io/spring-cloud-config/reference/server/environment-repository/jdbc-backend.html), so we can opt for this approach.

The only downside of Spring Cloud Config Server is that it is an extra microservice in the infrastructure. But since we decided to support a multi APP architecture, this is less relevant. This is not a problem for the developer experience, since we can make the apps work also when the config server is not available (`optional:configserver`), like explained [in this tutorial](https://www.baeldung.com/spring-cloud-configuration).

Currenlty all Geonetwork 5 configuration is managed in application Yamls. The idea is to start moving configurations away from this file (maybe keeping in the yaml only spring boot specific, database, elastic etc.?). If we want to proceed with the `optional:configserver`, it's a good idea to keep default values for relevant configurations in the yaml, otherwise we can remove them completely.

### What does this mean for other apps?

Almost zero impact on the code, all the settings are used like they are now in the application.yml. It's transparent where they are. Just need to configure them to use Spring Cloud Config Server and it's location. When the config server is configured with `optional:configserver`, the local yml is used when the server is not available, having zero impact.

### What is missing

Spring Cloud Config is read only, so we still need to implement an API to edit configurations, and eventually provide to a UI frontend only specific parts.
 
To evaluate if Spring Config Server may be a good option to keep also the translations i18n for backend and UI.

PascalLike:
From david video:

[...]

PR #168, opened 2026-05-04
Adding Spring Cloud Config based configuration

PascalLike:
# Spring Cloud Config Server to manage configuration

## Description
To have an alternative central way to manage apps configurations like https://github.com/geonetwork/geonetwork/issues/165, without rely on application yamls.
Allowing to change the configurations at runtime, notifing multiple apps.

This PR introduces:

- Spring Cloud Config Server and the configuration to use it in `src/apps/geonetwork`: This is configured to use a new db table `APP_CONFIGS` with configuration overrides. Spring Cloud Config Server has to stay private and not exposed on the internet. From code, parameters has to be used [like this](https://github.com/GeoCat/geonetwork/blob/configuration-on-spring-cloud-server/docs/manual/docs/Developer/spring-cloud-config-server.md#using-variables-in-code) 
- A new module `src/modules/configuration-management-api` (and shared module for domain logic `src/shared/configuration-management`) to draft [an API to access/manage configuration](https://github.com/GeoCat/geonetwork/blob/configuration-on-spring-cloud-server/docs/manual/docs/Developer/spring-cloud-config-server.md#runtime-configuration-api).

More details [here](https://github.com/GeoCat/geonetwork/blob/configuration-on-spring-cloud-server/docs/manual/docs/Developer/spring-cloud-config-server.md).

## Type of Change
Select one:
- [ ] Bug fix
- [X] New feature
- [ ] Documentation update
- [ ] Refactoring

## Related Issue
Closes/updates #156

## Approach
Discussed here #156

**This PR only introduces the configuration management mechanism based on Spring Cloud Config Server and an API for managing configurations. It lacks any security considerations, which should at least be considered before considering this development suitable for production.**

## How to Verify
1. Start postgres db (the same you will use for geonetwork!) 
2. (From src/apps) `mvn spring-boot:run -pl spring-cloud-config-server -Dspring-boot.run.jvmArguments="-DJDBC_DATABASE_URL=jdbc:postgresql://your-host:5432/geonetwork -DJDBC_DATABASE_USERNAME=geonetwork -DJDBC_DATABASE_PASSWORD=your_password"`
3. Start other components and geonetwork (`mvn spring-boot:run -pl geonetwork`)
4. API endpoints to check/change configurations are http://localhost:7979/geonetwork/doc/api/discover#gn5/tag/configuration-controller

## Checklist
- [X] All tests pass locally and in CI
- [X] Tests updated (if applicable)
- [X] Documentation updated (if applicable)


davidblasby:
@PascalLike - looks good.  I think this will make a big improvement.

The only thing I see is the implementation and (web) interfaces aren't separated.  There isn't much to the implementation so this is understandable. 

However,
modules - this is where the web interfaces (endpoints) should be defined (with openapi metadata)
shared - this is where the implementation should be

This is to allow other apps to decide if they want to bring in the endpoints or just the implementation.  However, in this case I'm not sure if that make too much sense since the individual components will likely just be bringing in spring-config?   I see a class like `ConfigurationUpdatedEvent` which would likely be needed by most of the other modules to get config change notifications?  

I think there could be stand-alone apps that don't have the config API endpoints (i.e. an app that's just ogcapi-records that only needs to be aware of the configuration (and config changes)).

Lets discuss and try to add a simple config to the ogcapi-records and something else to see if there are any impacts.

src/modules/configuration-management-api/src/main/java/org/geonetwork/configuration/ConfigurationController.java
```diff
@@ -0,0 +1,85 @@
...
+    @GetMapping
+    @Operation(
+            summary = "Get a configuration value",
+            description = "Retrieve a configuration parameter value from the Spring Environment using its key.")
+    public String get(
+            @Parameter(description = "The configuration parameter key (e.g., 'geonetwork.url')", required = true)
+                    @RequestParam
+                    String key) {
+        return configurationService.getConfiguration(key);
+    }
```
juanluisrp, line 66: `GET /api/configuration?key=` can leak secrets. `env.getProperty(key)` returns any Spring Environment property, including `spring.datasource.password`, OAuth client secrets, etc. The existing security config in `WebSecurityConfiguration.java` has "`**`" in the `permitAll()` list, so this endpoint is effectively public. Even with proper auth, returning arbitrary env values is dangerous.

Either remove this endpoint entirely, or restrict it to a known-safe key prefix (e.g., `geonetwork.*`) and ensure it is protected by an admin/operator role.

src/apps/spring-cloud-config-server/src/main/resources/schema.sql
```diff
@@ -0,0 +1,13 @@
...
+-- SPDX-FileCopyrightText: 2001 FAO-UN and others <geonetwork@osgeo.org>
+--
+-- SPDX-License-Identifier: GPL-2.0-or-later
+--
+CREATE TABLE IF NOT EXISTS APP_CONFIGS (
+  APP VARCHAR(255),
+  PROFILE VARCHAR(255),
+  LABEL VARCHAR(255),
+  CONFIG_PARAM VARCHAR(255),
+  CONFIG_VALUE VARCHAR(255),
```
juanluisrp: `CONFIG_VALUE VARCHAR(255)` truncates values silently. Connection strings, JSON blobs, or certificate thumbprints will be silently truncated at 255 chars, corrupting configuration. The Spring Cloud Config reference schema uses `VARCHAR(4096)` in the tests.

Increase to at least `VARCHAR(4096)` (or `TEXT`/`CLOB`).

src/shared/configuration-management/src/main/java/org/geonetwork/configuration/ConfigurationService.java
```diff
@@ -0,0 +1,100 @@
...
+
+    private final Environment env;
+    private final AppConfigRepository repository;
+    private final ApplicationEventPublisher eventPublisher;
+
+    @Value("${spring.application.name:GeoNetwork}")
+    private String defaultApp;
+
+    @Value("${spring.profiles.active:default}")
+    private String defaultProfile;
```
juanluisrp: This breaks with multiple active profiles. I don't know if it's a problem.
When more than one profile is active (common in this project, e.g. `default`,`jdbc`), this injects the raw comma-separated string "`default,jdbc`". The DB query `PROFILE=?` will then match no rows. The same issue affects `defaultApp` if `SPRING_APPLICATION_NAME` is ever a composite value.

Use `env.getActiveProfiles()` directly (taking the first element, or joining with `,` to match how Spring Cloud Config itself queries) instead of `@Value`

src/modules/configuration-management-api/src/main/java/org/geonetwork/configuration/ConfigurationInitializer.java
```diff
@@ -0,0 +1,48 @@
...
+
+            // Example default configuration
+            saveDefault("GeoNetwork", "default", "master", "geonetwork.api.configuration.enabled", "true");
+            saveDefault("GeoNetwork", "default", "master", "geonetwork.api.configuration.refresh-on-write", "true");
+            saveDefault("Frontend", "default", "master", "geonetwork.homepage.title", "GeoNetwork");
+            saveDefault("Frontend", "default", "master", "geonetwork.homepage.search.show", "false");
+
+            log.info("Configuration initialization complete.");
+
+            log.info("Configuration initialization complete.");
```
juanluisrp: Duplicated log line. Remove one.
```suggestion
            log.info("Configuration initialization complete.");
```

src/modules/configuration-management-api/src/main/java/org/geonetwork/configuration/ConfigurationInitializer.java
```diff
@@ -0,0 +1,48 @@
...
+    @Override
+    public void run(String... args) {
+        if (repository.count() == 0) {
+            log.info("Initializing APP_CONFIGS table with default values...");
+
+            // Example default configuration
+            saveDefault("GeoNetwork", "default", "master", "geonetwork.api.configuration.enabled", "true");
+            saveDefault("GeoNetwork", "default", "master", "geonetwork.api.configuration.refresh-on-write", "true");
+            saveDefault("Frontend", "default", "master", "geonetwork.homepage.title", "GeoNetwork");
+            saveDefault("Frontend", "default", "master", "geonetwork.homepage.search.show", "false");
```
juanluisrp: Is this example default configuration or the real default configuration that is going to be inserted in the production server? If the latter please update the comment.

src/modules/configuration-management-api/src/main/java/org/geonetwork/configuration/ConfigurationController.java
```diff
@@ -0,0 +1,85 @@
...
+     *
+     * @param config the configuration object
+     * @return success message
+     */
+    @PutMapping
+    @Operation(
+            summary = "Update a configuration parameter",
+            description =
+                    "Update an existing configuration parameter in the database. The context (app, profile, label) is automatically determined if not explicitly provided in the request body.")
+    public String update(@RequestBody AppConfig config) {
```
juanluisrp: `AppConfig` entity used directly as request body with no validation.

There's no `@Valid` annotation and no null-guard on the methods used in `configurationService.updateConfiguration` method. A `null` key will cause an opaque IllegalArgumentException or NPE deep in the service. A dedicated request DTO with `@NotBlank` on required fields would be cleaner and safer.

docs/manual/docs/Developer/spring-cloud-config-server.md
```diff
@@ -0,0 +1,136 @@
...
+## Database Structure
+The configuration is stored in the `APP_CONFIGS` table with the following structure:
+
+```sql
+CREATE TABLE APP_CONFIGS (
+  APP VARCHAR(255),           -- Application name (e.g., 'GeoNetwork')
+  PROFILE VARCHAR(255),       -- Spring profile (e.g., 'default', 'prod')
+  LABEL VARCHAR(255),         -- Optional label (e.g., 'master')
+  CONFIG_PARAM VARCHAR(255),  -- Configuration key (e.g., 'geonetwork.index.indexPrefix')
+  CONFIG_VALUE VARCHAR(255)   -- Configuration value
```
juanluisrp: If you update the configuration in the SQL update the doc to match.

juanluisrp (changes requested):
The secret-leaking GET endpoint and the `VARCHAR(255)` truncation are blocking issues to me. The multi-profile `@Value` injection is a silent correctness bug.

src/apps/spring-cloud-config-server/src/main/resources/schema.sql
```diff
@@ -0,0 +1,13 @@
...
+  PRIMARY KEY (APP, PROFILE, LABEL, CONFIG_PARAM)
+);
```
juanluisrp, line 14: I would add also a index here.

src/apps/spring-cloud-config-server/src/main/resources/schema.sql
juanluisrp: Checking the documentation at https://docs.spring.io/spring-cloud-config/reference/server/environment-repository/jdbc-backend.html should the `CONFIG_PARAM` and `CONFIG_VALUE` be named `key` and `value`?

src/modules/configuration-management-api/src/main/java/org/geonetwork/configuration/ConfigurationController.java
PascalLike: I changed the behavior as follows:

- Properties can only be modified with an admin profile.
- The admin can view all properties.
- Any other user, even one who is not logged in, can have read-only access only to properties that have APP_CONFIGS.INTERNAL set to false.
- Configurations that extend a value in YAML are created always with INTERNAL = false (these properties are used directly in the code via `@Value`)

src/apps/spring-cloud-config-server/src/main/resources/schema.sql
PascalLike: `CONFIG_VALUE` is now `TEXT`. 
`KEY` and `VALUE` are reserved words in H2 database and Postgres. In Postgres we could use the double quote, but let's keep things simple.

src/modules/configuration-management-api/src/main/java/org/geonetwork/configuration/ConfigurationInitializer.java
PascalLike: It was just for test. I commented the lines so we keep just them as example.

src/modules/configuration-management-api/src/main/java/org/geonetwork/configuration/ConfigurationController.java
PascalLike: Added validation.

src/shared/configuration-management/src/main/java/org/geonetwork/configuration/ConfigurationService.java
PascalLike: Followed your suggestion

PascalLike:
Followed the suggestion from @davidblasby to split the new module and creating a shared module with all domain logic.

Addressed feedbacks from @juanluisrp , answered inline. I think the security major concern is now addressed (see the [comment](https://github.com/geonetwork/geonetwork/pull/168#discussion_r3226418894) )  

Thank you very much for the useful review!

config/application.yml
```diff
@@ -90,6 +95,7 @@ spring:
       enabled: false
   config:
     import:
+      - optional:configserver:http://localhost:8888
```
davidblasby, line 98: might want this to be environment variable so its easier to inject into containers.

davidblasby:
It looks like "label" is a fundamental sub-slice of the data.  However, I'm not quite sure what it means.  Its not defined anywhere (the only help i see is its called "Configuration label ".  I thought this was a human-readable label about the config_param, but I don't think that right.  If its not a human-readable label please rename it and make sure it has javadocs defining what it means.

EDIT: looking at it deeper, this is part of spring config terminology.  However, it is confusing.

src/modules/configuration-management-api/src/main/java/org/geonetwork/configuration/ConfigurationController.java
```diff
@@ -0,0 +1,89 @@
...
+            description = "Retrieve a value from the Environment. Some parameters are visible to admin only.")
+    public String get(
+            @Parameter(description = "The configuration parameter key", required = true) @RequestParam String key) {
+        if (securityService.isAdmin()) {
+            return configurationService.getConfiguration(key);
+        } else {
+            if (!configurationService.canAccess(key)) {
+                throw new AccessDeniedException("Access to configuration parameter '" + key + "' is denied.");
+            }
+            return configurationService.getConfiguration(key);
```
davidblasby, line 64: This method (in the service) doesn't seem to access the database.  It doesn't seem to pay attention to defaultApp, defaultProfile, defaultLabel (which is used by db-based `#canAccess()`)

```
env.getProperty(key);
```

src/modules/configuration-management-api/src/main/java/org/geonetwork/configuration/ConfigurationController.java
```diff
@@ -0,0 +1,89 @@
...
+    @Operation(
+            summary = "Get a configuration value",
+            description = "Retrieve a value from the Environment. Some parameters are visible to admin only.")
+    public String get(
+            @Parameter(description = "The configuration parameter key", required = true) @RequestParam String key) {
+        if (securityService.isAdmin()) {
+            return configurationService.getConfiguration(key);
+        } else {
+            if (!configurationService.canAccess(key)) {
+                throw new AccessDeniedException("Access to configuration parameter '" + key + "' is denied.");
```
davidblasby, line 62: if the key doesn't exist, you get a 403.  It might want to give a 404 instead???

Here is the response:

{"timestamp":"2026-05-13T22:18:27.631+00:00","status":403,"error":"Forbidden","path":"/api/configuration"}

Looks like the exception message isn't being returned.

EDIT: I need to re-check this because of how the security was interacting.

src/modules/configuration-management-api/src/main/java/org/geonetwork/configuration/ConfigurationController.java
```diff
@@ -0,0 +1,89 @@
...
+
+    /**
+     * Get a configuration value from the Environment. Non-admins can only retrieve non-internal parameters that exist
+     * in the database.
+     */
+    @GetMapping
+    @Operation(
+            summary = "Get a configuration value",
+            description = "Retrieve a value from the Environment. Some parameters are visible to admin only.")
+    public String get(
```
davidblasby, line 56: This doesn't allow the user to specify app, profile, or label.

Note: canAccess() will use defaultApp, defaultProfile, defaultLabel

src/modules/configuration-management-api/src/main/java/org/geonetwork/configuration/ConfigurationController.java
```diff
@@ -0,0 +1,89 @@
...
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RequestParam;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+@RequestMapping("/api/configuration")
+@RequiredArgsConstructor
+@Validated
+@Tag(name = "Configuration", description = "API for managing application configuration parameters")
+public class ConfigurationController {
```
davidblasby: there's a missing `create` endpoint - although that might not be necessary.

davidblasby:
See: https://github.com/GeoCat/geonetwork/tree/_config2

I had to make changes to get this working.

davidblasby:
```
curl http://localhost:8888/config-server/jdbc/master | jq
  
{
  "name": "config-server",
  "profiles": [
    "jdbc"
  ],
  "label": "master",
  "version": null,
  "state": null,
  "propertySources": [
    {
      "name": "config-server-jdbc",
      "source": {
        "aaa": "value-aaabbb"
      }
    }
  ]
}
```

This endpoint is from Spring's `EnvironmentController#getEnvironment()`

davidblasby:
@PascalLike - I am going to stop looking at this now.  I am not sure if I understand it well enough to review it and I am probably just putting in garbage comments because I don't understand it!!!

With an addition like `#isInternal`, I am not sure if this is correctly leveraging the spring-cloud-config infrastructure.  For example, if you set `isInternal=true`, then the standard spring endpoint (cf. `EnvironmentController`):

```
 curl http://localhost:8888/config-server/jdbc/master
 ```

Will return the internal settings to everyone (no security check).

I believe that spring-cloud-config does some value layering (i.e. you can override values for different apps like you can in yml files), which I don't know if it will be with the gn `ConfigurationService` (i.e. `EnvironmentController#getEnvironment()` seems to be executing multiple sql queries).

PascalLike:
> @PascalLike - I am going to stop looking at this now. I am not sure if I understand it well enough to review it and I am probably just putting in garbage comments because I don't understand it!!!
> 
> With an addition like `#isInternal`, I am not sure if this is correctly leveraging the spring-cloud-config infrastructure. For example, if you set `isInternal=true`, then the standard spring endpoint (cf. `EnvironmentController`):
> 
> ```
>  curl http://localhost:8888/config-server/jdbc/master
> ```
> 
> Will return the internal settings to everyone (no security check).
> 
> I believe that spring-cloud-config does some value layering (i.e. you can override values for different apps like you can in yml files), which I don't know if it will be with the gn `ConfigurationService` (i.e. `EnvironmentController#getEnvironment()` seems to be executing multiple sql queries).

Thanks @davidblasby , I'm sorry if it was confusing. I changed the PR description to give more guidance.

Please also refer to the document available [here](https://github.com/GeoCat/geonetwork/blob/configuration-on-spring-cloud-server/docs/manual/docs/Developer/spring-cloud-config-server.md)

PR #179, opened 2026-07-10
Ogc api records facets and property mapping db backed configuration

### LINKED ISSUE #165: Use DB to configure Facets & properties

tanzeeladebont:
# OGC API Records Facets and Property-Mapping DB-backed Configuration (Issue #165)
## Description
Previously the OGC API Records facets and property-mapping configuration lived entirely in
`config/application-ogcapi-records.yml`. This change moves
the configuration into the database so it can be updated at runtime via a REST API.

## Type of Change
Select one:
- [ ] Bug fix
- [x] New feature
- [ ] Documentation update
- [ ] Refactoring

## Related Issue
Closes/updates # (optional):

## Checklist

### Database

Four tables are created by Liquibase changeset `00028-ogcapi-config.xml`. On first startup all rows are automatically seeded from the YAML config — no manual DB work needed.

**`ogcapi_property_mapping`** — one row, holds top-level settings and the update counter

| Column | Type | Description |
|---|---|---|
| `id` | `VARCHAR(255)` PK | Always `"property-mapping"` |
| `default_bucket_count` | `INTEGER` | Default number of facet buckets |
| `update_sequence` | `BIGINT` | Increments on every admin update |

**`ogcapi_field_mapping`** — one row per configured OGC API field

| Column | Type | Description |
|---|---|---|
| `id` | `BIGINT` PK | Auto-generated |
| `config_id` | `VARCHAR(255)` FK | → `ogcapi_property_mapping.id` |
| `sort_order` | `INTEGER` | Preserves field ordering |
| `ogc_property` | `VARCHAR(255)` | OGC property name exposed in the API |
| `elastic_property` | `VARCHAR(255)` | Dot-path into the Elasticsearch index |
| `index_record_property` | `VARCHAR(255)` | Dot-path into `IndexRecord` object |
| `type_override` | `VARCHAR(50)` | Optional type override (e.g. `LIST_STRING`) |
| `sort_field_suffix` | `VARCHAR(255)` | Suffix appended when used as a sort field |
| `is_sortable` | `BOOLEAN` | Expose in `/sortables` |
| `is_queryable` | `BOOLEAN` | Expose in `/queryables` |
| `title` | `VARCHAR(255)` | Human-readable label |
| `description` | `TEXT` | Human-readable description |
| `add_property_to_output` | `BOOLEAN` | Include value in record JSON output |

**`ogcapi_facet_config`** — one row per facet per field

| Column | Type | Description |
|---|---|---|
| `id` | `BIGINT` PK | Auto-generated |
| `field_mapping_id` | `BIGINT` FK | → `ogcapi_field_mapping.id` |
| `sort_order` | `INTEGER` | Preserves facet ordering within a field |
| `facet_name` | `VARCHAR(255)` | Name shown in the facets response |
| `facet_type` | `VARCHAR(50)` | `TERM`, `HISTOGRAM_FIXED_BUCKET_COUNT`, `HISTOGRAM_FIXED_INTERVAL`, or `FILTER` |
| `bucket_sorting` | `VARCHAR(50)` | `COUNT` or `VALUE` |
| `bucket_sorting_direction` | `VARCHAR(50)` | `ASCENDING` or `DESCENDING` |
| `bucket_count` | `INTEGER` | Max number of buckets |
| `minimum_document_count` | `INTEGER` | Buckets with fewer docs are hidden (default 1) |
| `number_bucket_interval` | `DOUBLE PRECISION` | Histogram interval for numeric fields |
| `calendar_interval_unit` | `VARCHAR(50)` | Histogram interval for date fields (`year`, `month`, …) |

**`ogcapi_filter_facet`** — one row per filter entry (only for `FILTER`-type facets)

| Column | Type | Description |
|---|---|---|
| `id` | `BIGINT` PK | Auto-generated |
| `facet_config_id` | `BIGINT` FK | → `ogcapi_facet_config.id` |
| `sort_order` | `INTEGER` | Preserves filter ordering within a facet |
| `filter_name` | `VARCHAR(255)` | Display name of the filter |
| `filter_equation_cql` | `TEXT` | CQL expression that defines the filter |

**Entity-Relationship Diagram**

```mermaid
erDiagram
    ogcapi_property_mapping {
        varchar id PK
        integer default_bucket_count
        bigint  update_sequence
    }

    ogcapi_field_mapping {
        bigint  id PK
        varchar config_id FK
        integer sort_order
        varchar ogc_property
        varchar elastic_property
        varchar index_record_property
        varchar type_override
        varchar sort_field_suffix
        boolean is_sortable
        boolean is_queryable
        varchar title
        text    description
        boolean add_property_to_output
    }

    ogcapi_facet_config {
        bigint  id PK
        bigint  field_mapping_id FK
        integer sort_order
        varchar facet_name
        varchar facet_type
        varchar bucket_sorting
        varchar bucket_sorting_direction
        integer bucket_count
        integer minimum_document_count
        double  number_bucket_interval
        varchar calendar_interval_unit
    }

    ogcapi_filter_facet {
        bigint  id PK
        bigint  facet_config_id FK
        integer sort_order
        varchar filter_name
        text    filter_equation_cql
    }

    ogcapi_property_mapping ||--o{ ogcapi_field_mapping : "has fields"
    ogcapi_field_mapping    ||--o{ ogcapi_facet_config   : "has facets"
    ogcapi_facet_config     ||--o{ ogcapi_filter_facet   : "has filters"
```

### New files

| File | Purpose |
|---|---|
| `src/shared/domain/.../OgcApiPropertyMapping.java` | JPA entity for `ogcapi_property_mapping` |
| `src/shared/domain/.../OgcApiFieldMapping.java` | JPA entity for `ogcapi_field_mapping` |
| `src/shared/domain/.../OgcApiFacetConfig.java` | JPA entity for `ogcapi_facet_config` |
| `src/shared/domain/.../OgcApiFilterFacet.java` | JPA entity for `ogcapi_filter_facet` |
| `src/shared/domain/.../repository/OgcApiPropertyMappingRepository.java` | JPA repository |
| `src/shared/ogcapi-records/.../OgcApiConfigChangedEvent.java` | Spring event fired on config update |
| `src/shared/ogcapi-records/.../OgcApiPropertyMappingService.java` | Service — seed, read, update config |
| `src/modules/ogcapi-records/.../OgcApiConfigController.java` | Admin REST endpoints |
| `src/shared/database/.../00028-ogcapi-config.xml` | Liquibase changeset (4 tables) |

### Modified files

| File | What changed |
|---|---|
| `OgcElasticFieldsMapperConfig.java` | `@Configuration` → `@Component` (removes CGLIB proxy so Jackson can serialize it) |
| `DynamicPropertiesFacade.java` | Loads config from `OgcApiPropertyMappingService` on startup; reloads on `OgcApiConfigChangedEvent` |
| `ElasticTypingSystem.java` | Added `refresh(newConfig)` to rebuild type maps; maps made `volatile` for thread safety |

---

## REST endpoints

Base path: `/api/ogcapi/configuration`

| Method | Path | Auth | Description |
|---|---|---|---|
| `GET` | `/api/ogcapi/configuration` | Admin | Return current config as JSON |
| `PUT` | `/api/ogcapi/configuration` | Admin | Replace config, increment sequence, reload in-memory cache |
| `GET` | `/api/ogcapi/configuration/update-sequence` | None | Return current update sequence counter |

The `/update-sequence` endpoint is unauthenticated so external containers can poll it to detect
changes without a message queue.

---

## How it works

```
YAML (bootstrap)
    ↓  first startup only — OgcApiPropertyMappingService @PostConstruct
ogcapi_property_mapping / ogcapi_field_mapping / ogcapi_facet_config / ogcapi_filter_facet
    ↓  every request
OgcApiPropertyMappingService.getConfig()   (entity graph → POJO)
    ↓  cached in-memory
DynamicPropertiesFacade  ←  OgcApiConfigChangedEvent (fired on PUT)
    ↓
facets / queryables / sortables endpoints
```

1. On startup, if no `property-mapping` row exists, the YAML config is mapped to the entity graph and inserted across all four tables.
2. `DynamicPropertiesFacade` loads config from the DB at startup via `@PostConstruct`.
3. When an admin calls `PUT /api/ogcapi/configuration`, the service clears the existing entity graph, rebuilds it from the new config, increments
   `update_sequence`, and fires `OgcApiConfigChangedEvent`.
4. `DynamicPropertiesFacade` listens for the event and reloads its cached config. It also calls
   `ElasticTypingSystem.refresh()` to rebuild the type maps.
5. All subsequent facet/queryable requests use the new config without a restart.

---

## Testing

**Unit tests**

```bash
mvn -pl src/shared/ogcapi-records test -Dtest=OgcApiPropertyMappingServiceTest
```

**Manual — with a running app (default port 7979, context path `/geonetwork`)**

```bash
# Check update sequence (no auth needed)
curl http://localhost:7979/geonetwork/api/ogcapi/configuration/update-sequence

# Sign in and save session cookie
curl -c /tmp/gn.cookies -X POST http://localhost:7979/geonetwork/api/user/signin \
  -d "username=admin&password=<password>"

# Get current config
curl -b /tmp/gn.cookies http://localhost:7979/geonetwork/api/ogcapi/configuration

# Update config (round-trip example — pipes GET response into PUT)
curl -b /tmp/gn.cookies http://localhost:7979/geonetwork/api/ogcapi/configuration \
  | curl -b /tmp/gn.cookies -X PUT \
      http://localhost:7979/geonetwork/api/ogcapi/configuration \
      -H "Content-Type: application/json" -d @-

# Confirm sequence incremented
curl http://localhost:7979/geonetwork/api/ogcapi/configuration/update-sequence
```

### DIFF
--- a/src/modules/ogcapi-records/ogcapi-records-implementation/src/main/java/org/geonetwork/ogcapi/records/OgcApiConfigController.java
+++ b/src/modules/ogcapi-records/ogcapi-records-implementation/src/main/java/org/geonetwork/ogcapi/records/OgcApiConfigController.java
@@ -1,0 +1,50 @@
+/*
+ * SPDX-FileCopyrightText: 2001 FAO-UN and others <geonetwork@osgeo.org>
+ * SPDX-License-Identifier: GPL-2.0-or-later
+ */
+package org.geonetwork.ogcapi.records;
+
+import io.swagger.v3.oas.annotations.Operation;
+import io.swagger.v3.oas.annotations.tags.Tag;
+import lombok.RequiredArgsConstructor;
+import org.geonetwork.ogcapi.service.configuration.OgcApiPropertyMappingService;
+import org.geonetwork.ogcapi.service.configuration.OgcElasticFieldsMapperConfig;
+import org.springframework.http.MediaType;
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.PutMapping;
+import org.springframework.web.bind.annotation.RequestBody;
+import org.springframework.web.bind.annotation.RequestMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+/** Admin endpoints for managing the OGC API Records facets and property-mapping configuration. */
+@RestController
+@RequestMapping("/api/ogcapi/configuration")
+@RequiredArgsConstructor
+@Tag(name = "OGC API Configuration", description = "Manage OGC API Records facets and property-mapping config")
+public class OgcApiConfigController {
+
+    private final OgcApiPropertyMappingService configService;
+
+    @Operation(summary = "Get the current OGC API property-mapping configuration")
+    @GetMapping(produces = MediaType.APPLICATION_JSON_VALUE)
+    // @PreAuthorize("hasRole('Administrator')")
+    public OgcElasticFieldsMapperConfig getConfiguration() {
+        return configService.getConfig();
+    }
+
+    @Operation(summary = "Replace the OGC API property-mapping configuration")
+    @PutMapping(consumes = MediaType.APPLICATION_JSON_VALUE, produces = MediaType.APPLICATION_JSON_VALUE)
+    //  @PreAuthorize("hasRole('Administrator')")
+    public OgcElasticFieldsMapperConfig updateConfiguration(@RequestBody OgcElasticFieldsMapperConfig newConfig) {
+        return configService.updateConfig(newConfig);
+    }
+
+    @Operation(
+            summary = "Get the current update sequence",
+            description = "Returns a counter incremented on every config update. "
+                    + "External containers can poll this to detect changes without a message queue.")
+    @GetMapping(path = "/update-sequence", produces = MediaType.APPLICATION_JSON_VALUE)
+    public long getUpdateSequence() {
+        return configService.getUpdateSequence();
+    }
+}
--- a/src/modules/ogcapi-records/ogcapi-records-implementation/src/main/java/org/geonetwork/ogcapi/records/OgcApiConfigController.java
+++ b/src/modules/ogcapi-records/ogcapi-records-implementation/src/main/java/org/geonetwork/ogcapi/records/OgcApiConfigController.java
@@ -39,13 +39,4 @@ public class OgcApiConfigController {
     public OgcElasticFieldsMapperConfig updateConfiguration(@RequestBody OgcElasticFieldsMapperConfig newConfig) {
         return configService.updateConfig(newConfig);
     }
-
-    @Operation(
-            summary = "Get the current update sequence",
-            description = "Returns a counter incremented on every config update. "
-                    + "External containers can poll this to detect changes without a message queue.")
-    @GetMapping(path = "/update-sequence", produces = MediaType.APPLICATION_JSON_VALUE)
-    public long getUpdateSequence() {
-        return configService.getUpdateSequence();
-    }
 }
--- a/src/modules/ogcapi-records/ogcapi-records-implementation/src/main/java/org/geonetwork/ogcapi/records/OgcApiConfigController.java
+++ b/src/modules/ogcapi-records/ogcapi-records-implementation/src/main/java/org/geonetwork/ogcapi/records/OgcApiConfigController.java
@@ -10,6 +10,7 @@ import lombok.RequiredArgsConstructor;
 import org.geonetwork.ogcapi.service.configuration.OgcApiPropertyMappingService;
 import org.geonetwork.ogcapi.service.configuration.OgcElasticFieldsMapperConfig;
 import org.springframework.http.MediaType;
+import org.springframework.security.access.prepost.PreAuthorize;
 import org.springframework.web.bind.annotation.GetMapping;
 import org.springframework.web.bind.annotation.PutMapping;
 import org.springframework.web.bind.annotation.RequestBody;
@@ -27,14 +28,14 @@ public class OgcApiConfigController {
 
     @Operation(summary = "Get the current OGC API property-mapping configuration")
     @GetMapping(produces = MediaType.APPLICATION_JSON_VALUE)
-    // @PreAuthorize("hasRole('Administrator')")
+    @PreAuthorize("hasRole('Administrator')")
     public OgcElasticFieldsMapperConfig getConfiguration() {
         return configService.getConfig();
     }
 
     @Operation(summary = "Replace the OGC API property-mapping configuration")
     @PutMapping(consumes = MediaType.APPLICATION_JSON_VALUE, produces = MediaType.APPLICATION_JSON_VALUE)
-    //  @PreAuthorize("hasRole('Administrator')")
+    @PreAuthorize("hasRole('Administrator')")
     public OgcElasticFieldsMapperConfig updateConfiguration(@RequestBody OgcElasticFieldsMapperConfig newConfig) {
         return configService.updateConfig(newConfig);
     }
--- a/src/shared/ogcapi-records/src/main/java/org/geonetwork/ogcapi/service/configuration/OgcApiPropertyMappingService.java
+++ b/src/shared/ogcapi-records/src/main/java/org/geonetwork/ogcapi/service/configuration/OgcApiPropertyMappingService.java
@@ -7,7 +7,6 @@ package org.geonetwork.ogcapi.service.configuration;
 import jakarta.annotation.PostConstruct;
 import java.util.ArrayList;
 import java.util.List;
-import java.util.concurrent.atomic.AtomicInteger;
 import lombok.RequiredArgsConstructor;
 import lombok.extern.slf4j.Slf4j;
 import org.geonetwork.domain.OgcApiFacetConfig;
@@ -77,9 +76,9 @@ public class OgcApiPropertyMappingService {
                 .build());
 
         entity.getFields().clear();
-        AtomicInteger fieldOrder = new AtomicInteger(0);
+        int fieldOrder = 0;
         for (OgcElasticFieldMapperConfig fieldConfig : newConfig.getFields()) {
-            OgcApiFieldMapping fieldEntity = toFieldEntity(fieldConfig, entity, fieldOrder.getAndIncrement());
+            OgcApiFieldMapping fieldEntity = toFieldEntity(fieldConfig, entity, fieldOrder++);
             entity.getFields().add(fieldEntity);
         }
         entity.setDefaultBucketCount(newConfig.getDefaultBucketCount());
@@ -165,9 +164,9 @@ public class OgcApiPropertyMappingService {
                 .defaultBucketCount(yaml.getDefaultBucketCount())
                 .updateSequence(1L)
                 .build();
-        AtomicInteger fieldOrder = new AtomicInteger(0);
+        int fieldOrder = 0;
         for (OgcElasticFieldMapperConfig fieldConfig : yaml.getFields()) {
-            OgcApiFieldMapping fieldEntity = toFieldEntity(fieldConfig, entity, fieldOrder.getAndIncrement());
+            OgcApiFieldMapping fieldEntity = toFieldEntity(fieldConfig, entity, fieldOrder++);
             entity.getFields().add(fieldEntity);
         }
         return entity;
@@ -192,10 +191,10 @@ public class OgcApiPropertyMappingService {
                 .description(fieldConfig.getDescription())
                 .addPropertyToOutput(fieldConfig.getAddPropertyToOutput())
                 .build();
-        AtomicInteger facetOrder = new AtomicInteger(0);
+        int facetOrder = 0;
         if (fieldConfig.getFacetsConfig() != null) {
             for (OgcFacetConfig facetConfig : fieldConfig.getFacetsConfig()) {
-                OgcApiFacetConfig facetEntity = toFacetEntity(facetConfig, fieldEntity, facetOrder.getAndIncrement());
+                OgcApiFacetConfig facetEntity = toFacetEntity(facetConfig, fieldEntity, facetOrder++);
                 fieldEntity.getFacets().add(facetEntity);
             }
         }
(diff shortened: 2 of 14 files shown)


src/modules/ogcapi-records/ogcapi-records-implementation/src/main/java/org/geonetwork/ogcapi/records/OgcApiConfigController.java
```diff
@@ -0,0 +1,50 @@
...
+@RequestMapping("/api/ogcapi/configuration")
+@RequiredArgsConstructor
+@Tag(name = "OGC API Configuration", description = "Manage OGC API Records facets and property-mapping config")
+public class OgcApiConfigController {
+
+    private final OgcApiPropertyMappingService configService;
+
+    @Operation(summary = "Get the current OGC API property-mapping configuration")
+    @GetMapping(produces = MediaType.APPLICATION_JSON_VALUE)
+    // @PreAuthorize("hasRole('Administrator')")
```
PascalLike: Please uncomment this and following ones, it may be useful for testing, but better do not have it in the code. And please use `@PreAuthorize("@securityService.isAdmin()")` (geonetwork specific profile) instead of `@PreAuthorize("hasRole('Administrator')")`

src/shared/ogcapi-records/src/main/java/org/geonetwork/ogcapi/service/indexConvert/dynamic/DynamicPropertiesFacade.java
```diff
@@ -27,14 +31,27 @@
...
     ExtraElasticPropertiesService extraElasticPropertiesService;
 
+    private OgcElasticFieldsMapperConfig config;
+
+    @PostConstruct
+    void loadConfig() {
+        config = configService.getConfig();
+    }
+
+    @EventListener
```
PascalLike: Replace `@EventListener` with `@TransactionalEventListener(phase = TransactionPhase.AFTER_COMMIT)` to guarantee the cache only reloads after the transaction has successfully committed.

src/shared/ogcapi-records/src/main/java/org/geonetwork/ogcapi/service/indexConvert/dynamic/DynamicPropertiesFacade.java
```diff
@@ -27,14 +31,27 @@
...
-    private OgcElasticFieldsMapperConfig config;
+    private OgcApiPropertyMappingService configService;
 
     @Autowired
     ElasticTypingSystem elasticTypingSystem;
 
     @Autowired
     ExtraElasticPropertiesService extraElasticPropertiesService;
 
+    private OgcElasticFieldsMapperConfig config;
```
PascalLike: Declare the `config` field as `volatile`so it's thread safe

```java
private volatile OgcElasticFieldsMapperConfig config;
```

src/modules/ogcapi-records/ogcapi-records-implementation/src/main/java/org/geonetwork/ogcapi/records/OgcApiConfigController.java
```diff
@@ -0,0 +1,50 @@
...
+    //  @PreAuthorize("hasRole('Administrator')")
+    public OgcElasticFieldsMapperConfig updateConfiguration(@RequestBody OgcElasticFieldsMapperConfig newConfig) {
+        return configService.updateConfig(newConfig);
+    }
+
+    @Operation(
+            summary = "Get the current update sequence",
+            description = "Returns a counter incremented on every config update. "
+                    + "External containers can poll this to detect changes without a message queue.")
+    @GetMapping(path = "/update-sequence", produces = MediaType.APPLICATION_JSON_VALUE)
```
PascalLike: That seems a smart approach, how it should work on other nodes? This maybe a good alternative to adopting a message broker. I created an issue to keep track of the options (https://github.com/geonetwork/geonetwork/issues/180) If you don't mind, can you please add this idea of the `/update-sequence` in the issue? It worth to be considered as an alternative!

PascalLike (commented):
Thanks @tanzeeladebont, it was wonderful reviewing this PR: perfect description, well detailed design approach, a very clear and complete description of how to perform tests. 🌟  

I've tested it and it work as expected! Added some comments in the code, please give a look.

Please also review if the documentation here https://github.com/geonetwork/geonetwork/tree/dev/docs/manual/docs/OGCAPI-Records is aligned. 
Especially [this](https://github.com/geonetwork/geonetwork/blob/dev/docs/manual/docs/OGCAPI-Records/config.md)

src/shared/ogcapi-records/src/main/java/org/geonetwork/ogcapi/service/configuration/OgcApiPropertyMappingService.java
```diff
@@ -0,0 +1,248 @@
...
+
+    @PostConstruct
+    void seedFromYamlIfAbsent() {
+        if (repository.existsById(CONFIG_ID)) {
+            return;
+        }
+        OgcApiPropertyMapping entity = buildEntityFromYaml(yamlConfig);
+        repository.save(entity);
+        log.info("Seeded OGC API property-mapping config from YAML into DB.");
+    }
```
jahow, line 47: Why are we seeding from the YAML file if we can initiate a config using Liquibase? It feels strange to have these two configuration systems in parallel

src/modules/ogcapi-records/ogcapi-records-implementation/src/main/java/org/geonetwork/ogcapi/records/OgcApiConfigController.java
jahow: I can see how this can be useful, but who actually asked for this? Feels like YAGNI should dictate to remove this altogether

jahow (commented):
I was surprised to see this PR popup in my feed. Could anyone please point me to where the need to have runtime changes on the configuration is expressed? In GN4 the index configuration is also in a static ~YAML~ JSON file and I have never heard of it being a problem. Storing configuration in the database has strong implications for the future of the app.

Edit: ok I found the related issue by @davidblasby that this addresses. Still, I'd really be interested in knowing more about the benefits brought by this change. Again, I think having a configuration as a static file makes things so much simpler, especially in a multi-node context.

Thanks

davidblasby:
Olivia,

This has to do with the ogcapi records configuration. I.e. facets config and how the index is mapped to the ogc-defined JSON (mostly in the properties section). It also does a lot behind the scenes (I.e. add to queriables, sortables, CQL, and the openapi (swagger) doc).  I expect a lot can be added in the future too (computational fields, conversions, multi-index-property fields etc…). 

This is something that would be modified over time, as well as be deployment-specific.  I expect the next step will be to add a simple gui on top that would allow someone to modify/view the config.  This will be great for a multi-node deployment system too (esp during an upgrade). 

This is a great addition that will be very useful!

@tanzeeladebont - great PR!

jahow:
> Olivia,
> 
> This has to do with the ogcapi records configuration. I.e. facets config and how the index is mapped to the ogc-defined JSON (mostly in the properties section). It also does a lot behind the scenes (I.e. add to queriables, sortables, CQL, and the openapi (swagger) doc). I expect a lot can be added in the future too (computational fields, conversions, multi-index-property fields etc…).
> 
> This is something that would be modified over time, as well as be deployment-specific. I expect the next step will be to add a simple gui on top that would allow someone to modify/view the config. This will be great for a multi-node deployment system too (esp during an upgrade).
> 
> This is a great addition that will be very useful!
> 
> @tanzeeladebont - great PR!

Thanks David, I understand that but I still struggle to see the added value in having this modifiable at runtime. Specially with multiple nodes! having static files dictate the structure of the OGC API endpoint sounds so much easier in many aspects. Don't you agree?

davidblasby:
Olivia,

I like gui-based config - like in GN4 and in geoserver. I also like config in the database. Much friendlier than having to manage/edit config files, especially for multiple nodes. For the admin, even being able to see the config is a big win. 

I see configuring the ogcapi output (and how everything is connected) as a big win for any installation. Perhaps our difference is that I see this as something the admin would use to make their install “right” instead of something totally static. Even something like adding facets for end user searchability and data grouping!

jahow:
> Olivia,
> 
> I like gui-based config - like in GN4 and in geoserver. I also like config in the database. Much friendlier than having to manage/edit config files, especially for multiple nodes. For the admin, even being able to see the config is a big win.
> 
> I see configuring the ogcapi output (and how everything is connected) as a big win for any installation. Perhaps our difference is that I see this as something the admin would use to make their install “right” instead of something totally static. Even something like adding facets for end user searchability and data grouping!

Thanks for the explanation, I think this PR has seen enough misplaced discussions. I can see the benefit of having config in the database; if only for the fact that it makes deployment of the app easier by using the same docker image all the time :)

src/modules/ogcapi-records/ogcapi-records-implementation/src/main/java/org/geonetwork/ogcapi/records/OgcApiConfigController.java
tanzeeladebont: Thank you so much for creating this issue. I have added my opinion there . To me this one seems like a quick solution ready to be merged while we take our time to create the infrastructure and implement a message broker. If you also agree with me, should we consider keeping this endpoint ?

src/modules/ogcapi-records/ogcapi-records-implementation/src/main/java/org/geonetwork/ogcapi/records/OgcApiConfigController.java
tanzeeladebont: Thank you for catching this. I was supposed to uncomment it before I commit it.

PascalLike (approved):
Thanks @tanzeeladebont , it looks good to me

Issue #189, opened 2026-07-30
OGC API - Records: CQL NOT is ignored, negated filters return the un-negated result set


Ali-D-Akbar:
OGC API - Records ignores CQL negation: `NOT (...)` returns the same records as the un-negated predicate, with HTTP 200. `<>` behaves as `=`. Nothing on the client side can detect this, which makes it worse than an error response.

Found while wiring a catalogue search UI to the OGC API - Records backend, where every filter offers include and exclude. Exclude currently returns the opposite record set.

## Environment

- `dev` at 79f4011, run with `spring-boot:run`
- Postgres and Elasticsearch from `docker/docker-compose-dbs.yml` (sample data, `gn-records`)
- Collection `3bef299d-cf82-4033-871b-875f6936b2e2`, 30 records, of which 28 are `resourceType = dataset`, 11 are `spatialRepresentationType = vector`, 1 is `orgForResource = DREAL`

## Repro

```
GET /geonetwork/ogcapi-records/collections/{id}/items?limit=1&filter=<expr>
```

| `filter=` | `numberMatched` | expected |
| --- | --- | --- |
| *(none)* | 30 | 30 |
| `resourceType = 'dataset'` | 28 | 28 |
| `NOT (resourceType = 'dataset')` | **28** | 2 |
| `NOT (resourceType = 'map')` | **0** | 30 |
| `resourceType <> 'dataset'` | **28** | 2 |
| `NOT resourceType = 'dataset'` | **28** | 2 |
| `spatialRepresentationType = 'vector'` | 11 | 11 |
| `NOT (spatialRepresentationType = 'vector')` | **11** | 19 |
| `orgForResource = 'DREAL'` | 1 | 1 |
| `NOT (orgForResource = 'DREAL')` | **1** | 29 |
| `(resourceType = 'dataset') AND (NOT (orgForResource = 'DREAL'))` | **0** | 27 |

The pattern is that the negation is dropped and the inner predicate is evaluated. `NOT (resourceType = 'map')` returning 0 is the clearest tell: it yields exactly the `resourceType = 'map'` result.

Alternative syntaxes are not a workaround, all HTTP 500:

- `resourceType != 'dataset'`
- `NOT (resourceType IN ('dataset'))`
- `resourceType NOT IN ('dataset')`
- `resourceType IS NULL`
- `NOT resourceType IS NULL`

`NOT (resourceType LIKE 'dataset')` returns 200 with 28, the same drop-the-negation behaviour.

[...]

PR #191, opened 2026-08-03
Fix: Ogc api records ignores cql negation

tanzeeladebont:
# Fix: OGC API Records ignores CQL negation (`NOT`, `<>`)

Fixes [geonetwork/geonetwork#189](https://github.com/geonetwork/geonetwork/issues/189).

## Symptom

`GET /geonetwork/ogcapi-records/collections/{id}/items?filter=...` silently dropped negation:

| `filter=` | returned | expected |
| --- | --- | --- |
| `NOT (resourceType = 'dataset')` | same as `resourceType = 'dataset'` | the inverse set |
| `resourceType <> 'dataset'` | same as `resourceType = 'dataset'` | the inverse set |

No error was returned (HTTP 200), so clients had no way to detect that the filter was silently wrong.

## Root cause

`ImprovedCqlFilter2Elastic` (`src/shared/ogcapi-records/src/main/java/org/geonetwork/ogcapi/service/cql/ImprovedCqlFilter2Elastic.java`)
is a GeoTools `AbstractFilterVisitor` that walks a parsed CQL filter and pushes Elasticsearch `Query`
objects onto a shared stack (`must`/`should`/`term`/`range`, etc.). It had overrides for `PropertyIsEqualTo`,
`PropertyIsLike`, range comparisons, `PropertyIsBetween`, `And`, and `Or` — but **no override for `Not` or
`PropertyIsNotEqualTo`**.

Without an override, GeoTools falls back to `AbstractFilterVisitor`'s default behavior:

- **`Not`**: the default implementation just visits the child filter and discards the wrapper. Since our
  visitor builds queries by pushing onto a shared stack as a side effect, the child's query still ended up
  on the stack — as if the `NOT` had never been there.
- **`PropertyIsNotEqualTo` (`<>`)**: the default delegates to a shared expression-visitor mechanism that our
  subclass doesn't actually wire up for this path, so the clause pushed *nothing* onto the stack. A
  standalone `<>` filter silently evaluated to "match everything"; inside `AND`/`OR` it could pop the wrong
  sibling's query off the stack and corrupt the whole compound filter.

## Fix

Added the two missing overrides in `ImprovedCqlFilter2Elastic`:

```java
@Override
public Object visit(Not filter, Object extraData) {
    filter.getFilter().accept(this, extraData);
    var negatedQuery = (Query) stack.pop();
    stack.push(Query.of(q -> q.bool(b -> b.mustNot(negatedQuery))));
    return this;
}

@Override
public Object visit(PropertyIsNotEqualTo filter, Object extraData) {
    var equalsQuery = buildEqualsQuery(filter, extraData);
    stack.push(Query.of(q -> q.bool(b -> b.mustNot(equalsQuery))));
    return this;
}
```

`PropertyIsEqualTo`'s query-building logic was extracted into a shared `buildEqualsQuery(...)` helper so
`<>` reuses the exact same term-query construction, just wrapped in `bool.must_not`.

### Multi-valued fields

The issue also asked to confirm behavior on multi-valued fields (`resourceType`, `tags`, `format`,
`orgForResource`, `spatialRepresentationType`): excluding value `V` should mean "no element of the array
equals `V`", not "some element differs from `V`".

No special-casing was needed. Elasticsearch `term` queries on array/keyword-list fields already match if
*any* element equals the value, so wrapping that same query in `bool.must_not` correctly negates to "no
element equals the value" — exactly the semantics the issue asked for.

## Tests

Added cases to `ImprovedCqlFilter2ElasticTest.java` covering:

- `dave <> 'hi'` → `bool.must_not` wrapping a `term` query
- `not (dave = 'hi')` → `bool.must_not` wrapping a `term` query
- `(dave = 'hi') and (not (dave2 = 'bob'))` → `bool.must` with one `term` and one nested `bool` clause

### DIFF
--- a/src/shared/ogcapi-records/src/main/java/org/geonetwork/ogcapi/service/cql/ImprovedCqlFilter2Elastic.java
+++ b/src/shared/ogcapi-records/src/main/java/org/geonetwork/ogcapi/service/cql/ImprovedCqlFilter2Elastic.java
@@ -73,6 +73,28 @@ public class ImprovedCqlFilter2Elastic extends AbstractFilterVisitor {
 
     @Override
     public Object visit(PropertyIsEqualTo filter, Object extraData) {
+        stack.push(buildEqualsQuery(filter, extraData));
+        return this;
+    }
+
+    @Override
+    public Object visit(PropertyIsNotEqualTo filter, Object extraData) {
+        var equalsQuery = buildEqualsQuery(filter, extraData);
+        stack.push(Query.of(q -> q.bool(b -> b.mustNot(equalsQuery))));
+        return this;
+    }
+
+    @Override
+    public Object visit(Not filter, Object extraData) {
+        filter.getFilter().accept(this, extraData);
+
+        var negatedQuery = (Query) stack.pop();
+
+        stack.push(Query.of(q -> q.bool(b -> b.mustNot(negatedQuery))));
+        return this;
+    }
+
+    private Query buildEqualsQuery(BinaryComparisonOperator filter, Object extraData) {
         checkFilterExpressionsInBinaryComparisonOperator(filter);
 
         filter.getExpression1().accept(expressionVisitor, extraData);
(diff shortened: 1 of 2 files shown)


Ali-D-Akbar (commented):
Thanks for the quick fix @tanzeeladebont. We've tested this on our side and negation works correctly now.

One concern... And it's in config rather than your diff. `resourceTitleObject` maps to `resourceTitleObject.default`, which is analyzed text, so a `term` query can never match a full title. `=` returns 0 today. Now that negation works, `NOT (resourceTitleObject = '...')` returns every record instead.

I could be wrong but the `.keyword` subfield already exists:

```diff
 - ogcProperty: resourceTitleObject
-   elasticProperty: resourceTitleObject.default
+   elasticProperty: resourceTitleObject.default.keyword
```

`elasticProperty` also feeds facets and sorting. So worth a check before applying. Not blocking.

Issue #186, opened 2026-07-27
LIKE with % wildcards fails with a double URL-decoding error server-side.


tanzeeladebont:
LIKE with % wildcards fails with a double URL-decoding error server-side.

PR #193, opened 2026-08-07
Fix: LIKE with `%` wildcards fails with a double URL-decoding error

tanzeeladebont:
# Fix: LIKE with `%` wildcards fails with a double URL-decoding error

Fixes [geonetwork/geonetwork#186](https://github.com/geonetwork/geonetwork/issues/186).

## Symptom

`GET /ogcapi-records/collections/{id}/items?filter=name%20LIKE%20%27%25foo%25%27`
(i.e. CQL `name LIKE '%foo%'`) failed server-side with:

```
java.lang.IllegalArgumentException: URLDecoder: Illegal hex characters in escape (%) pattern - Error at index 1 in: "fo"
```

The request never reached CQL parsing — it failed before that, inside query-parameter processing.

## Where the request goes wrong, step by step

1. **Client sends** the CQL text `name LIKE '%foo%'` as a `filter` query parameter, percent-encoded
   so the literal `%` becomes `%25`:
   `filter=name%20LIKE%20%27%25foo%25%27`.

2. **Spring MVC / the servlet container decodes it once (correctly).** The generated controller
   interface binds it as a plain `@RequestParam`:

   `ogcapi-records-openapi-autogen/.../generated/CollectionsApi.java:548`
   ```java
   @RequestParam(value = "filter", required = false) @Nullable String filter,
   ```

   By the time `OgcapiCollectionsApiController.getRecords(...)` receives `filter`, Tomcat has
   already turned it back into the plain string `name LIKE '%foo%'`. This is correct and complete —
   no further decoding is needed or expected.

3. **The controller passes `filter` straight through**, unmodified, into
   `queryBuilder.buildFromRequest(...)`.

4. **`QueryBuilder` decodes it a second time — the bug.** In
   `src/shared/ogcapi-records/src/main/java/org/geonetwork/ogcapi/service/querybuilder/QueryBuilder.java`,
   the method ran:

   ```java
   var _filter = (filter == null) ? null : URLDecoder.decode(filter, StandardCharsets.UTF_8);
   result.setFilter(_filter);
   ```

   `URLDecoder.decode` treats any `%` in its input as the start of a percent-encoded byte and expects
   two hex digits after it. Given the already-decoded string `name LIKE '%foo%'`, it looks at `%fo`
   and tries to parse `fo` as hex — `o` isn't a valid hex digit, so it throws `IllegalArgumentException`
   before the string ever reaches `org.geotools.filter.text.cql2.CQL` / `ImprovedCqlFilter2Elastic`.

   The same redundant `URLDecoder.decode(...)` call also existed for `datetime` (line 86) and for
   every "extra queryable" parameter value in `injectExtraFromRequest` (line 132). Those don't
   typically contain a `%`, which is why the bug only surfaced for `LIKE` filters — but a value that
   happens to contain a `%xx` sequence that *is* valid hex (e.g. `%41` → `A`) would silently be
   corrupted there instead of throwing, which is arguably worse since it fails silently.

   `parameterMap` (the source for the "extra queryable" values) comes from
   `HttpServletRequest.getParameterMap()` in the controller — also already decoded by the container,
   so the same reasoning applies to all three call sites.

## Fix

Removed all three redundant `URLDecoder.decode(...)` calls in `QueryBuilder.java` and use the
values as received — they're already correctly decoded by Spring MVC by the time they reach this
method:

```java
// before
var _filter = (filter == null) ? null : URLDecoder.decode(filter, StandardCharsets.UTF_8);
result.setFilter(_filter);

// after
result.setFilter(filter);
```

Same pattern applied to `datetime` and to `result.addProp(...)` in `injectExtraFromRequest`. The
now-unused `java.net.URLDecoder` / `java.nio.charset.StandardCharsets` imports were removed too.

## Why this is safe

Nothing upstream of `QueryBuilder` ever passes it a raw, still-encoded string — every caller binds
these values via Spring `@RequestParam` or `HttpServletRequest.getParameterMap()`, both of which the
servlet container decodes exactly once. There was never a legitimate case where a second decode was
needed; it was a leftover/defensive call that only manifested as a bug once a value legitimately
contained a `%`.

## Tests

Added `testFilterWithLikeWildcardIsNotDoubleDecoded` in
`src/shared/ogcapi-records/src/test/java/org/geonetwork/ogcapi/service/querybuilder/QueryBuilderTest.java`,
asserting that a filter value of `name LIKE '%foo%'` passes through `buildFromRequest` unchanged
instead of throwing or being corrupted.

No test previously exercised a `%`-containing value through this path, which is why the bug went
unnoticed by CI (existing integration tests URL-encode their CQL before sending, but none of them
use `LIKE` or a raw `%`).

## A trap when testing this manually

Since the server now only decodes the `filter` parameter once (the correct amount), the value must
also only be percent-encoded **once** on the way in. If you take an already-encoded string like
`name%20LIKE%20%27%25foo%25%27` and paste it into a tool that itself percent-encodes the value
(a Swagger UI field, `curl --data-urlencode`, a browser address bar that auto-encodes), you get a
second layer of encoding on the wire. The server decodes that one layer and is left with the
still-encoded text `name%20LIKE%20%27%25foo%25%27`, which fails CQL lexing with the same shape of
error (`Encountered: "%" ... Parsing : name%20LIKE...`) — this looks like the original bug but is a
client-side double-encoding mistake, not a server issue.

Always give the tool the **plain** CQL text and let it encode once:

```bash
curl -G 'http://localhost:8080/geonetwork/ogcapi-records/collections/{id}/items' \
  --data-urlencode "filter=name LIKE '%foo%'"
```

## Follow-up: real clients still send it double-encoded, and we don't control them

OGC API - Records is a public API. Some real callers (third-party/open-source consumers, not just
GeoNetwork's own UI) genuinely send the `filter` value percent-encoded twice, for the same reason
described in the trap above — and unlike our own frontend, we can't just go fix their code. Simply
requiring "encode it exactly once" isn't enforceable on every caller, so the *server* needs to
tolerate both cases.

### The fix: try first, decode-and-retry only on failure

`CqlToElasticSearch` (`src/shared/ogcapi-records/src/main/java/org/geonetwork/ogcapi/service/cql/CqlToElasticSearch.java`)
is where the `filter` string is actually parsed as CQL (`CQL.toFilter(...)`). CQL syntax validity is
a discriminator we can use instead of guessing: a correctly single-encoded value always parses
successfully on the first attempt, so it's never touched further; a value that's still one layer
over-encoded fails to parse as CQL (that's the `%` lexical error), and *only* in that case do we try
decoding it once more and re-parsing:

```java
Filter parseCql(String cql) throws CQLException {
    try {
        return CQL.toFilter(cql);
    } catch (CQLException firstAttemptFailure) {
        String decodedAgain;
        try {
            decodedAgain = URLDecoder.decode(cql, StandardCharsets.UTF_8);
        } catch (IllegalArgumentException notPercentEncoded) {
            throw firstAttemptFailure;
        }
        if (decodedAgain.equals(cql)) {
            throw firstAttemptFailure;
        }
        try {
            return CQL.toFilter(decodedAgain);
        } catch (CQLException secondAttemptFailure) {
            throw firstAttemptFailure;
        }
    }
}
```

Both `create(OgcApiQuery)` and `create(String cql)` now call `parseCql(...)` instead of
`CQL.toFilter(...)` directly. If the retry doesn't help (genuinely invalid CQL, or a decode that
doesn't change the string), the *original* parse error is thrown, so error messages stay accurate
for real syntax mistakes.

### DIFF
--- a/src/shared/ogcapi-records/src/test/java/org/geonetwork/ogcapi/service/querybuilder/QueryBuilderTest.java
+++ b/src/shared/ogcapi-records/src/test/java/org/geonetwork/ogcapi/service/querybuilder/QueryBuilderTest.java
@@ -35,6 +35,7 @@ public class QueryBuilderTest {
     List<String> ids;
     List<String> externalids;
     List<String> sortby;
+    String filter;
     Map<String, String[]> parameterMap;
     List<String> advancedFacets;
 
@@ -72,6 +73,7 @@ public class QueryBuilderTest {
         ids = Arrays.asList("id1", "id2");
         externalids = Arrays.asList("ex-id1", "ex-id2");
         sortby = Arrays.asList("sort-p1", "sort-p2");
+        filter = null;
         parameterMap = new LinkedHashMap<>();
     }
 
@@ -134,6 +136,20 @@ public class QueryBuilderTest {
                 query.getAdvancedFacets().get(0).getSorting());
     }
 
+    /**
+     * A `filter` (or `datetime`, or queryable value) arriving at {@link QueryBuilder} is already URL-decoded by Spring
+     * MVC / the servlet container. Values that contain a literal `%` (e.g. an OGC CQL `LIKE` wildcard) must be passed
+     * through as-is, not decoded again - otherwise `%foo%` is misread as a percent-encoded byte and either throws or is
+     * corrupted.
+     */
+    @Test
+    public void testFilterWithLikeWildcardIsNotDoubleDecoded() throws Exception {
+        filter = "name LIKE '%foo%'";
+        var query = buildSampleQuery();
+
+        assertEquals("name LIKE '%foo%'", query.getFilter());
+    }
+
     public OgcApiQuery buildSampleQuery() throws Exception {
         return queryBuilder.buildFromRequest(
                 collectionId,
@@ -146,7 +162,7 @@ public class QueryBuilderTest {
                 ids,
                 externalids,
                 sortby,
-                null,
+                filter,
                 "cql2-text",
                 null,
                 parameterMap,
--- a/src/modules/ogcapi-records/ogcapi-records-integration-tests/src/test/java/org/geonetwork/dynamicproperties/DynamicPropertiesTest.java
+++ b/src/modules/ogcapi-records/ogcapi-records-integration-tests/src/test/java/org/geonetwork/dynamicproperties/DynamicPropertiesTest.java
@@ -13,8 +13,6 @@ import com.fasterxml.jackson.databind.node.TextNode;
 import com.google.common.base.Ascii;
 import com.google.common.collect.Iterables;
 import com.google.common.collect.Ordering;
-import java.net.URLEncoder;
-import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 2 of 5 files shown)


src/shared/ogcapi-records/src/test/java/org/geonetwork/ogcapi/service/querybuilder/QueryBuilderTest.java
```diff
@@ -134,6 +136,20 @@ public void testWithAdvancedFacets() throws Exception {
...
+     * MVC / the servlet container. Values that contain a literal `%` (e.g. an OGC CQL `LIKE` wildcard) must be passed
+     * through as-is, not decoded again - otherwise `%foo%` is misread as a percent-encoded byte and either throws or is
+     * corrupted.
+     */
+    @Test
+    public void testFilterWithLikeWildcardIsNotDoubleDecoded() throws Exception {
+        filter = "name LIKE '%foo%'";
+        var query = buildSampleQuery();
+
+        assertEquals("name LIKE '%foo%'", query.getFilter());
```
PascalLike, line 150: Very minor thing: can this be `assertEquals(filter, query.getFilter());` ?

